PR #1, opened 2024-05-29
Add instruction and struct definitions

joncinque:
#### Problem

Before getting started on the staking program, we need to agree on an interface, ie. structs and instructions.

#### Summary of changes

Going off the work at https://github.com/paladin-bladesmith/paladin-rewards-program/pull/1, this adds the basics for the staking program.

After thinking about it some more, I came to the conclusion that we can neatly separate the staking portion from the governance portion. The governance program just needs to interpret the stake and config accounts from this program to figure out things like voting power, but that's the only hard link between the two.

The stake program can grant slash and config authority to the governance, and then the governance can vote on instructions to the stake program.

The biggest question that came out for the rewards program is around the handling of staking rewards: I think we'll need to track and store them separately from the holder rewards, otherwise it's difficult to figure out what proportion of rewards should go to which stake account. I'll flag those in the PR.


src/instruction.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// 0. `[]` Config account
+    /// 1. `[writable]` Stake account
+    /// 2. `[writable]` SOL staking rewards account
+    ///    (TODO for discussion: we need a way to also track total staking rewards
+    ///    so stakers can know their allotted proportion of staking rewards,
+    ///    which are separate from the holder rewards. This means that the distribution
+    ///    logic in the Rewards program *also* needs to update some running total
+    ///    of staking rewards. I couldn't come up with a way to combine these two,
+    ///    since the proportion allocated to the different groups is *not* meant
+    ///    to be fixed forever.)
```
joncinque: Here: I think we need to have some way of tracking stake rewards. `total_rewards` at https://github.com/paladin-bladesmith/paladin-rewards-program/pull/1/files#diff-87619baf236177a7a9fc38150961151b7583b6f2f082ade75915f3ac4ca49c41R75 assumes that the distribution between holders / stakers / leader / piggybank will never change. If we want it to ever change, we need to keep the running total of holder and staker rewards separately.

And then I'm not sure where the staker rewards should be stored, but I'm open to ideas!

src/instruction.rs
```diff
@@ -0,0 +1,185 @@
+//! Program instructions
+
+use solana_program::clock::UnixTimestamp;
+
+/// Instructions supported by the staking program
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum StakeInstruction {
+    /// Creates Stake config account which controls staking parameters
+    ///
+    /// 0. `[writable]` Stake config account
```
edgarpavlovsky: Nit: I don't have strong preference on which one we use, but can we align on one way to represent writable accounts across our programs? The stake program has `[writable]` and rewards has `[w]` right now

src/instruction.rs
```diff
@@ -0,0 +1,185 @@
...
+    ///   Must be the token account on the stake config account
+    /// 5. `[]` Token program
+    /// 6.. Extra accounts required for the transfer hook
+    ///
+    /// Instruction data: amount of tokens to stake, as a little-endian u64
+    StakeTokens(u64),
+
+    /// Deactivate staked tokens for the validator.
+    ///
+    /// Only one deactivation may be in-flight at once, so if this is called
```
edgarpavlovsky, line 59: Is there an attack vector here? ie if someone continuously calls deactivation

It sounds like no, because this ix isn't permissionless

src/instruction.rs
```diff
@@ -0,0 +1,185 @@
...
+    ///
+    /// Reduces the total voting power for the stake account and the total staked
+    /// amount on the system.
+    ///
+    /// NOTE: This instruction is permissionless, so anybody can finish
+    /// deactivating someone's tokens, preparing them to be withdrawn.
+    ///
+    /// 0. `[writable]` Stake config account
+    /// 1. `[writable]` Validator stake account
+    InactivateStake,
```
edgarpavlovsky, line 79: why do we need deactivation and deactivation finalization to be two separate ixs?

src/instruction.rs
edgarpavlovsky: this would be nice to find a solution for -- it feels like something that is likely to be up for at least a discussion of changing in the future

can we leverage a config + multi-vault architecture? where the config determines the split at a given point in time, and the vault distributes based on that config on a regular basis

src/instruction.rs
joncinque: Yeah sure, happy to go with either!

src/instruction.rs
joncinque: Correct, the authority must sign this, so if they reset their own deactivation, that's their fault. we could add a `force` flag to this instruction, but that can be done from the CLI / UI

src/instruction.rs
joncinque: It's mostly to make the accounting as clear and simple as possible. After a validator deactivates some portion of stake, they need to wait the full cooldown time until it's actually inactive and able to be withdrawn. During that cooldown time, the tokens are still staked and on the hook for any slashes.

Once that time passes, however, the tokens are inactive, and the new state of those tokens must be reflected in the system as a whole. For example, we need to update the `total_staked_amount` to subtract the inactive tokens from the pool. We don't have this issue in the native staking program because the runtime does a bunch of calculations at epoch boundaries for us. We don't have that luxury here. 

Without this field, a malicious validator can deactivate tokens, but never withdraw them, so they're still getting their full rewards and vote weight. And then at any time, when it's convenient for them (ie. there's an upcoming slash on them), they can withdraw those tokens.

src/instruction.rs
edgarpavlovsky: let's go with `[w]` for simplicity?

src/instruction.rs
edgarpavlovsky: ah gotcha -- good call, this makes sense!

.gitignore
buffalojoec, line 1: Do you want `Cargo.lock` ignored? And you've got two `target`s now.

src/instruction.rs
```diff
@@ -0,0 +1,185 @@
...
+use solana_program::clock::UnixTimestamp;
+
+/// Instructions supported by the staking program
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum StakeInstruction {
+    /// Creates Stake config account which controls staking parameters
+    ///
+    /// 0. `[writable]` Stake config account
+    /// 1. `[]` Slash authority, may be a PDA for a governance program
+    /// 2. `[]` Config authority, may be a PDA for a governance program
```
buffalojoec, line 12: Do these two need to be passed in as accounts, or can they be keys in the input data?

src/instruction.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// 3. `[]` Stake Token Mint
+    /// 4. `[]` Stake Token Vault, to hold all staked tokens.
+    ///   Must be empty and owned by PDA with seeds `['token-owner', stake_config]`
+    InitializeConfig {
```
buffalojoec, line 16: This instruction is permissionless. Can there be only one stake config? Is the stake config a PDA with some deterministic derivation?

src/instruction.rs
```diff
@@ -0,0 +1,185 @@
...
+        cooldown_time: UnixTimestamp,
+        max_deactivation_basis_points: u16,
+    },
+
+    /// Initializes stake account data for a validator.
+    ///
+    /// NOTE: Anybody can create the stake account for a validator. The authority
+    /// defaults to the validator withdraw authority.
```
buffalojoec: "Defaults" or "is always set to"?

src/instruction.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// 0. `[]` Stake config account
+    /// 1. `[writable]` Validator stake account
+    ///     * PDA seeds: ['stake', validator, config_account]
+    /// 2. `[]` Validator vote account
+    /// 3. `[]` System program
+    InitializeStake,
+
+    /// Stakes tokens with the given config.
+    ///
+    /// Limited to the current amount of SOL staked to the validator.
```
buffalojoec, line 36: I think I'm a little confused on this. It looks like this instruction enables staking of some mint to the validator, but it's limited by amount of _SOL_ somehow? Can you maybe explain to me a bit more about this instruction?

src/instruction.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// 1. `[writable]` Validator stake account
+    ///     * PDA seeds: ['stake', validator, config_account]
+    /// 2. `[writable]` Token Account
+    /// 3. `[signer]` Owner or delegate of the token account
+    /// 4. `[]` Validator vote account
+    /// 3. `[]` Stake Token Mint
+    /// 4. `[]` Stake Token Vault, to hold all staked tokens.
+    ///   Must be the token account on the stake config account
+    /// 5. `[]` Token program
+    /// 6.. Extra accounts required for the transfer hook
```
buffalojoec, line 52: Ahh interesting. So just for my understanding, the act of staking PAL tokens is effectively a transfer, which triggers the transfer hook to the Rewards program?

src/instruction.rs
buffalojoec, line 79: TIL this is a word. 😅

src/state.rs
```diff
@@ -0,0 +1,58 @@
...
+    solana_program::clock::UnixTimestamp, solana_program::pubkey::Pubkey,
+    spl_pod::optional_keys::OptionalNonZeroPubkey,
+};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum AccountType {
+    Uninitialized,
+    Config,
+    Stake,
+}
```
buffalojoec, line 13: Just curious; I don't see this used anywhere yet. Something akin to Token-2022 account state?

src/state.rs
```diff
@@ -0,0 +1,58 @@
...
+pub struct Config {
+    /// After a deactivation, defines the number of seconds that must pass before
+    /// the stake is inactive and able to be withdrawn
+    cooldown_time_secs: u64,
+    /// The maximum proportion that can be deactivated at once, given as basis
+    /// points (1 / 10,000)
+    max_deactivation_basis_points: u16,
+    /// Total number of tokens delegated to the system. Since anyone can transfer
+    /// tokens into the vault without passing through the program, this number
+    /// is maintained independently.
```
buffalojoec, line 25: There is still the transfer hook though, right? Maybe I'm not understanding, but we could still capture those transfers.

Maybe you mean this is specifically the tokens that have gone through the program and are effectively delegated/staked?

src/instruction.rs
buffalojoec: I have a few questions on this, but I'm thinking this is a good topic for tomorrow's call?

buffalojoec (commented):
Looking good! Mostly nits and clarifying questions!

src/instruction.rs
joncinque: They don't need to be passed in as accounts, but it can be better since you potentially dedupe the pubkey when compiling the instruction, and also get free validation that it's a well-formed pubkey. I lean towards doing it this way, but I can be swayed

src/instruction.rs
joncinque: There can be many stake configs, which would allow us to reuse the program for things like lockups

src/instruction.rs
joncinque: If you create a stake account for someone else, the authority will default to the validator withdraw authority. The validator may then set it to any other key. So it won't *always* be set to the withdraw authority. I can clarify, "For new stake accounts, the authority defaults..." -- what do you think?

src/instruction.rs
joncinque: For sure! The concept is that any validator cannot have more PAL staked than their amount of SOL staked. So for example, if I have 1,000 SOL (10^12 lamports) delegated to me on the network, I cannot stake more than a total of 1,000 PAL (10^12 nano-PAL).

src/instruction.rs
joncinque: Yep, any time we transfer PAL, we'll need all those extra accounts!

src/instruction.rs
joncinque: Haha sorry, I had a hard time coming up with a good word, so any suggestions are *very* welcome

src/state.rs
joncinque: It's essentially just a small account discriminator, but we should probably use `spl-discriminate` for these

.gitignore
joncinque: Nice catch, will fixup

src/state.rs
joncinque: In the governance program, to figure out the outcome of a vote, we need to know what proportion of stake has said "yes" or "no". The stake account stores the voting power of an individual staker, but we need to know how much voting power is part of the network.

To simplify things, we might think to simply read the amount of tokens in the PAL staker escrow. However, if we do that, someone could maliciously just transfer tokens directly into the escrow account to fudge the proportions.

For example, if there are 100 PAL tokens in the escrow, and you have a 51 PAL staked, you could immediately win any vote. But if I transfer 300 PAL tokens into the escrow without going through the staking procedure, then it would be impossible for the governance to ever resolve votes, since we could only get up to 100 PAL in voting power, and the system things there's 400 PAL in total voting power.

So to get around that, we store the total amount staked here

src/instruction.rs
joncinque: Yeah definitely, let's cover this

src/instruction.rs
buffalojoec: The only reason I lean toward providing them as inputs in the instruction data is because you're paying for compute to map the account infos into the VM and translate them into the account slice when you don't have to. In either case we're spending some compute on interpreting the public keys, but at least we can omit the rest of the metadata and the account's inner data.

I'd say up to you on how careful we want to be about compute!

src/instruction.rs
buffalojoec: Would the validator be using the `SetAuthority` instruction to do this reassign? If so, I'd say this instruction can explicitly say it's "always initialized to" or something like that, since this instruction itself isn't capable of assigning another authority.

src/instruction.rs
buffalojoec: Ok got it, thanks! Pegged stablecoin PAL. 😁

src/instruction.rs
buffalojoec: Ehh works for me honestly!

src/state.rs
buffalojoec: Up to you on the `spl-discriminate` bit. Saving 7 bytes isn't bad though!

src/state.rs
buffalojoec: Ok gotcha, thanks for the explanation!

src/instruction.rs
joncinque: Since initialize is only called once on a config, I don't think slightly higher compute usage should be a problem in exchange for some more clarity and safety.

src/instruction.rs
joncinque: works for me!

src/instruction.rs
buffalojoec: Sounds good, I think I'll lean into this with the rewards program's `InitializeStakerRewards` as well, then.

src/state.rs
febo: Might be easier to use `spl-discriminate` since the types are 8-byte aligned already.

src/state.rs
```diff
@@ -0,0 +1,58 @@
...
+    token_amount_delegated: u64,
+    /// Authority that can modify any elements in the config
+    config_authority: OptionalNonZeroPubkey,
+    /// Authority that can slash any stake
+    slash_authority: OptionalNonZeroPubkey,
+    /// Token account storing all stake
+    vault_token_account: Pubkey,
```
febo, line 32: Should this be the `mint` instead? The token account could have a deterministic derivation (e.g. , an ATA) and we probably need to validate the mint on stake instructions.

joncinque: This is just meant to be the token account -- but yeah, we'll need to do some validation on it during init, ie. to make sure the account is owned by the proper authority and doesn't have any extra extensions on it. And init also passes in the mint to do any checks on it too.

src/state.rs
joncinque: can we leave it for follow-up work?

joncinque:
This should be ready to go on my side!

buffalojoec (approved):
Lgtm, ship it!

PR #3, opened 2024-06-13
Use project template

febo:
This PR updates the repository to use the `create-solana-program` template plus adds:
1. Shank annotations to instructions
2. Empty processor for each instruction
3. Kinobi configuration
4. Generates IDL + JS and Rust clients

I am using the `spl-pod` crate for the `Config` account fields, which means that the account is "hybrid" since not all types are fully zero-copy. It was needed since there is an `u16` field on the account; also, I added an account discriminator (`u8`) and a bump (`u8`) for the vault signer. We could "bundle" these values together into a `data: [u8; 8]` field to have everything aligned to 8 bytes and add accessor methods for these 3 fields; or we could replace them and make the account 8 bytes aligned; or we could leave as it is.

I am not sure that we need to use `OptionalNonZeroPubkey` for the `slash_authority` on the `Config` account – would there be a case where slash would be disable by setting this authority to `None`?


buffalojoec:
Hey @febo, I'm happy to review this, but what do you think about taking a little bit more of an incremental approach?
See my PR at https://github.com/paladin-bladesmith/paladin-rewards-program/pull/8.

I think maybe we do one PR to set up the repository template - without touching the program - and then a new PR for Shank/IDL-gen, and then finally one or more PRs to port over the tests.

What do you think?

febo:
> Hey @febo, I'm happy to review this, but what do you think about taking a little bit more of an incremental approach? See my PR at [paladin-bladesmith/paladin-rewards-program#8](https://github.com/paladin-bladesmith/paladin-rewards-program/pull/8).
> 
> I think maybe we do one PR to set up the repository template - without touching the program - and then a new PR for Shank/IDL-gen, and then finally one or more PRs to port over the tests.
> 
> What do you think?

Yeah, happy to do that. I will take the opportunity to fix the template and then use the updated one. I will make all my PRs draft for the moment until I "port" them.

PR #7, opened 2024-06-14
Use project template

febo:
This PR applied the `create-solana-project` template to the repository. The existing interface is moved to `program/src` without any modification.

* [Remove file](https://github.com/paladin-bladesmith/stake-program/pull/7/commits/cfc1e979c5903d4f085133e26fde2f56c4adbab6)
* [Remove boilerplate text](https://github.com/paladin-bladesmith/stake-program/pull/7/commits/7d280dffa6865f43a2e12735ce51a3bab0b3f3f8)


.github/workflows/main.yml
```diff
@@ -0,0 +1,196 @@
...
+          node: ${{ env.NODE_VERSION }}
+          solana: ${{ env.SOLANA_VERSION }}
+      - name: Cache cargo dependencies
+        uses: actions/cache@v4
+        with:
+          path: ${{ env.CARGO_CACHE }}
+          key: ${{ runner.os }}-cargo-programs-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: ${{ runner.os }}-cargo-programs
+      - name: Build programs
+        run: pnpm programs:build
```
buffalojoec, line 38: I think we want the `bpf-entrypoint` feature here, right?

```suggestion
        run: pnpm programs:build --features bpf-entrypoint
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,196 @@
...
+        uses: actions/cache@v4
+        with:
+          path: ${{ env.CARGO_CACHE }}
+          key: ${{ runner.os }}-cargo-program-tests-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-program-tests
+            ${{ runner.os }}-cargo-programs-${{ hashFiles('**/Cargo.lock') }}
+            ${{ runner.os }}-cargo-programs
+      - name: Test programs
+        run: pnpm programs:test
```
buffalojoec, line 72: Same here?

```suggestion
        run: pnpm programs:test --features bpf-entrypoint
```

LICENSE
buffalojoec, line 1: We should check with @edgarpavlovsky before adding any license to these repositories.

README.md
buffalojoec, line 1: Up to you, but I canned all of this "getting started with the generated code" README stuff.

program/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+pub mod instruction;
+pub mod state;
+
+solana_program::declare_id!("PStake1111111111111111111111111111111111111");
```
buffalojoec, line 4: ???

scripts/generate-clients.mjs
```diff
@@ -0,0 +1,80 @@
...
+        k.constantPdaSeedNodeFromString("utf8", "counter"),
+        k.variablePdaSeedNode(
+          "authority",
+          k.publicKeyTypeNode(),
+          "The authority of the counter account"
+        ),
+      ],
+    },
+  })
+);
```
buffalojoec, line 34: I think all of these "counter" configs can get dropped, since they're not for the stake program. If we need any configs like these for the program, we can add them later!

buffalojoec (commented):
Lgtm, I left just a few small comments!

I'm curious why this didn't trigger CI. I also don't think it will pass, since all of the tests are outfitted to test the demo "counter" program. I had to actually scrub all of the tests and anything related to "counter" from the configs, etc.

.github/workflows/main.yml
febo: Yes, I was planning in fixing the CI in a separate PR, once it is setup on the repo. CI won't run on the branch until there is a workflow on main.

.github/workflows/main.yml
febo: Same as previous comment.

LICENSE
febo: Good call, I removed the file that comes from the template.

README.md
febo: Sounds good, I would do the same.

program/src/lib.rs
febo: This is just a temporary id, not a real one.

scripts/generate-clients.mjs
febo: My plan is to fix all the client generation in a separate PR, including adding a basic Kinobi configuration.

febo:
> Lgtm, I left just a few small comments!
> 
> I'm curious why this didn't trigger CI. I also don't think it will pass, since all of the tests are outfitted to test the demo "counter" program. I had to actually scrub all of the tests and anything related to "counter" from the configs, etc.

Thanks for the review. Addressed/replied your comments.

CI won't run from the branch until there is a workflow on main first.

buffalojoec (approved):
Ship it!

rustfmt.toml
```diff
@@ -0,0 +1,10 @@
+max_width = 100
+reorder_imports = true
+reorder_modules = true
```
joncinque, line 3: Is there a reason to specify these? They're all the default values according to https://rust-lang.github.io/rustfmt/?version=v1.6.0

joncinque (approved):
Just a little comment, but it sounds like the plan is to merge this as is and then remove stuff as needed, so I'll approve

rustfmt.toml
febo: Oh, nice. These are coming from the template, so we should fix it there as well. I will remove them once I enable `nightly` fmt.

PR #8, opened 2024-06-15
Add shank annotation

febo:
This PR adds Shank annotations to the instruction `enum` to generate an IDL using `pnpm generate:idls` (the IDL has been upddated as part of this PR).

This is a stacked PR – the relevant commit is:
* [Add shank annotation](https://github.com/paladin-bladesmith/stake-program/commit/12f8b25dace19dcfa543cd7846af4c7a0381fc07)

* [Add missing comments](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/86cb3e82747d02b5876f1336c18c939072fb0e6c)
* [Use unixtimestamp type](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/ffe7f0b811a75ee78202d57bbc91e8727cde2b98)
* [Fix seeds order](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/94ef372f41498fb7ef4aa958d7c498e26b182bb4)
* [Remove borsh](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/6f64bd351f064944f5df3da59494a458b2965a17)
* [Fix account ordering](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/23d37f99ca50f666c642ee8ee5fe83b1fd81ebfe)
* [Rename to vault](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/fbd70ee1afd9130d00cbe46af5f48b51d4d305af)
* [Rename instruction enum](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/bdf40ae6f167ca84ad9845e4e0ab312b9167abd9)

New commits based on 2nd review:
* [Rename to token program](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/45437b2a2e9a58d9c266f9db38675256cb1381fc)
* [Add missing seeds and spaces](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/836231713cf18522441ede1d0ef55bff09664ea8)
* [Rename to destination token account](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/6d24bcb8289227a57b55f9c9cee4b5ba161a6345)
* [Rename to source token account](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/beffa3fd55af6bd793b06e0792a46720ccddf84f)
* [Rename to token account authority](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/e65fe67bb6fbc2fe585f6a7536fa02d2b5aeefa5)
* [Add missing seeds](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/430029cf1f6cedcb7d40c2779df53ff2d4000c20)
* [Fix account description](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/6a6f378c0009b2cbd21eb372e7024628bdf52e11)
* [Rename to system program](https://github.com/paladin-bladesmith/stake-program/pull/8/commits/01a6a147490c13cdda11600579b22eae54843c70)


program/src/instruction.rs
```diff
@@ -1,184 +1,373 @@
...
+#[rustfmt::skip]
+#[derive(BorshDeserialize, BorshSerialize, Clone, Debug, ShankContext, ShankInstruction)]
+pub enum Instruction {
+    /// Creates Stake config account which controls staking parameters.
+    #[account(
+        0,
+        writable,
+        name = "config",
+        desc = "Stake config account"
+    )]
```
buffalojoec, line 14: Why drop the documentation? Does the macro attribute add it to the variant for you? If not, we should keep the doc comments that outline the instruction's accounts, etc.

program/src/instruction.rs
```diff
@@ -1,184 +1,373 @@
...
-    ///
-    /// 0. `[w]` Stake config account
-    /// 1. `[]` Slash authority, may be a PDA for a governance program
-    /// 2. `[]` Config authority, may be a PDA for a governance program
-    /// 3. `[]` Stake Token Mint
-    /// 4. `[]` Stake Token Vault, to hold all staked tokens.
-    ///   Must be empty and owned by PDA with seeds `['token-owner', stake_config]`
+/// Enum defining all instructions in the Stake program.
+#[rustfmt::skip]
+#[derive(BorshDeserialize, BorshSerialize, Clone, Debug, ShankContext, ShankInstruction)]
```
buffalojoec: Is Borsh required here?

program/src/instruction.rs
```diff
@@ -1,184 +1,373 @@
...
+        desc = "Stake token mint"
+    )]
+    #[account(
+        4,
+        name = "vault_token",
+        desc = "Stake token vault"
+    )]
     InitializeConfig {
-        cooldown_time: UnixTimestamp,
+        cooldown_time_seconds: u64,
```
buffalojoec: Does this work with the IDL if it's of type `UnixTimestamp`? Only asking because I think this change should be separated from the IDL PR.

program/src/instruction.rs
```diff
@@ -1,184 +1,373 @@
...
+    )]
+    #[account(
+        1,
+        name = "config_authority",
+        desc = "Config authority"
+    )]
+    #[account(
+        2,
+        name = "slash_authority",
+        desc = "Slash authority"
```
buffalojoec: These are flipped around. Might not matter since the processor isn't implemented, but just calling out!

program/src/instruction.rs
```diff
@@ -1,184 +1,373 @@
...
+    )]
+    #[account(
+        3,
+        name = "mint",
+        desc = "Stake token mint"
+    )]
+    #[account(
+        4,
+        name = "vault_token",
```
buffalojoec: I think the name `vault_token` could be confusing to people. They might think it's also a mint. What about `token_vault` or just `vault`?

program/src/instruction.rs
```diff
@@ -1,184 +1,373 @@
...
+    #[account(
+        0,
+        name = "config",
+        desc = "Stake config account"
+    )]
+    #[account(
+        1,
+        writable,
+        name = "stake",
+        desc = "Validator stake account (pda of `['stake', config, validator]`)"
```
buffalojoec: The last two seeds here are backwards. The original comment said `['stake', validator, config_account]`.

program/src/instruction.rs
```diff
@@ -1,184 +1,373 @@
...
+        6,
+        writable,
+        name = "vault_token",
+        desc = "Stake token Vault"
+    )]
+    #[account(
+        7,
+        name = "spl_token_program",
+        desc = "SPL Token 2022 program"
+    )]
```
buffalojoec: I'm not totally sure on this, but we might not want to make this specifically Token-2022. It could just be "a token program", in case people wanted to use other token programs in the future.

program/src/instruction.rs
```diff
@@ -1,184 +1,373 @@
...
+        3,
+        name = "vault_authority",
+        desc = "Vault authority"
+    )]
+    #[account(
+        4,
+        writable,
+        name = "vault_token",
+        desc = "Vault token account"
+    )]
```
buffalojoec: It looks like you jumbled up the order of these, too. Is this on purpose?

program/src/instruction.rs
```diff
@@ -1,184 +1,373 @@
...
+        5,
+        writable,
+        name = "destination_token",
+        desc = "Destination token account"
+    )]
+    #[account(
+        6,
+        name = "spl_token_program",
+        desc = "SPL Token program"
+    )]
```
buffalojoec: This one seems more generalized than the previous instruction which requires "Token-2022" specifically.

program/src/instruction.rs
```diff
@@ -1,184 +1,373 @@
...
+        2,
+        writable,
+        name = "vault_token",
+        desc = "Vault token account"
+    )]
+    #[account(
+        3,
+        name = "vault_authority",
+        desc = "Vault authority"
+    )]
```
buffalojoec: Same comment here about the account ordering being mixed up from the original comments.

program/src/instruction.rs
```diff
@@ -1,184 +1,373 @@
...
+        name = "system_program",
+        desc = "System program account"
+    )]
     DistributeRewards(u64),
 }
 
 /// Enum defining all authorities in the program
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub enum Authority {
+#[derive(BorshDeserialize, BorshSerialize, Clone, Debug, PartialEq, Eq)]
```
buffalojoec: Same comment about Borsh.

program/src/instruction.rs
```diff
@@ -1,184 +1,373 @@
...
+        writable,
+        name = "config",
+        desc = "Stake config account"
+    )]
+    #[account(
+        1,
+        writable,
+        signer,
+        name = "payer",
+        desc = "Reward payer"
```
buffalojoec: Also flipped.

buffalojoec (commented):
Looking good. I noticed a lot of accounts out of order from the original comments, so that's most of my feedback here. Was this intentional?

program/src/instruction.rs
```diff
@@ -1,184 +1,373 @@
...
-    /// 0. `[w]` Stake config account
-    /// 1. `[]` Slash authority, may be a PDA for a governance program
-    /// 2. `[]` Config authority, may be a PDA for a governance program
-    /// 3. `[]` Stake Token Mint
-    /// 4. `[]` Stake Token Vault, to hold all staked tokens.
-    ///   Must be empty and owned by PDA with seeds `['token-owner', stake_config]`
+/// Enum defining all instructions in the Stake program.
+#[rustfmt::skip]
+#[derive(BorshDeserialize, BorshSerialize, Clone, Debug, ShankContext, ShankInstruction)]
+pub enum Instruction {
```
buffalojoec: Can we keep this named `StakeInstruction`?

program/src/instruction.rs
febo: Added the missing comments. I did not add the "accounts" comments since these ones are the same as the Shank tags. Did you mean to leave those as well? I usually add them on the processor, but happy to have them here are well.

The generated code will have the accounts comments generated based on the Shank tags.

program/src/instruction.rs
febo: No, only if we are using borsh to deserialize the instruction data – since we are not, I will remove the `Borsh*` derives.

program/src/instruction.rs
febo: I reverted to `UnixTimestamp` on this PR.

The type does not work out-of-the-box with Kinobi. I will address this on the "generate clients" PR.

program/src/instruction.rs
febo: The order does not matter if you are using the `Context` derive, since you access accounts by name. I inverted them since so the `config_authority` comes right after  the `config` account, but this is totally cosmetic – I can switch it back if you prefer.

program/src/instruction.rs
febo: Good catch, thanks!

program/src/instruction.rs
febo: Yes, totally cosmetic. I prefer grouping the accounts, so everything related to `stake` is close together, then everything related to `vault` is close together, etc. Happy to revert it back to the original if that is better.

program/src/instruction.rs
febo: Hmmm, how would that work? Like, we need the transfer hook for this to work, right? So if we keep it generic than it "looks" like you could use SPL token.

program/src/instruction.rs
febo: Replied on the previous comment about this, but I definitely need to be consistent once we agree on an approach.

program/src/instruction.rs
febo: Please see comment above about account ordering.

program/src/instruction.rs
febo: Removed.

program/src/instruction.rs
febo: Same as previous, in this case `payer` stays "closer" to `system_program`.

program/src/instruction.rs
febo: Done.

program/src/instruction.rs
febo: Renamed to `vault`.

program/src/instruction.rs
febo: Reverted back to the original order.

program/src/instruction.rs
febo: Reverted back to the original order.

program/src/instruction.rs
febo: Reverted back to the original order.

program/src/instruction.rs
febo: Reverted back to the original order.

febo:
> Looking good. I noticed a lot of accounts out of order from the original comments, so that's most of my feedback here. Was this intentional?

It was intentional, but I reverted back to the original order.

buffalojoec:
> It was intentional, but I reverted back to the original order.

Cool, thanks. If you wanted to reorder them according to your preference I think that's fine, it just wasn't clear when I reviewed. I also have my own cosmetic preference to ordering accounts :)

program/src/instruction.rs
buffalojoec: Yeah you're right we do need the extensions, but typically we've been treating Token-2022's functionality as an "interface", just like Token-2022 implements (and extends) the SPL Token interface.

It's not really a perfect pattern right now, but I'd say let's just go with "Token program". The instruction should be compatible with any token program that implements, say "Transfer Hook".

program/src/instruction.rs
```diff
@@ -1,184 +1,420 @@
...
+        3,
+        writable,
+        name = "vault",
+        desc = "Vault token account"
+    )]
+    #[account(
+        4,
+        name = "vault_authority",
+        desc = "Vault authority"
+    )]
```
joncinque: This is my own lack of knowledge with shank -- does it allow you to specify PDA seeds anywhere? Either way, could we add that info here?

program/src/instruction.rs
```diff
@@ -1,184 +1,420 @@
...
+    #[account(
+        6,
+        name = "vault_authority",
+        desc = "Vault authority"
+    )]
+    #[account(
+        7,
+        name = "mint",
+        desc="Stake token mint"
```
joncinque: micro-nit: will rustfmt fix the missing spaces?

program/src/instruction.rs
```diff
@@ -1,184 +1,420 @@
...
+        5,
+        signer,
+        name = "stake_authority",
+        desc = "Stake authority"
+    )]
+    #[account(
+        6,
+        name = "vault_authority",
+        desc = "Vault authority"
+    )]
```
joncinque: Can you add the PDA info here?

program/src/instruction.rs
```diff
@@ -1,184 +1,420 @@
...
+        4,
+        signer,
+        name = "stake_authority",
+        desc = "Stake authority"
+    )]
+    #[account(
+        5,
+        name = "vault_authority",
+        desc = "Vault authority"
+    )]
```
joncinque: Same here: can you add the PDA seeds?

program/src/instruction.rs
```diff
@@ -1,184 +1,420 @@
...
+    #[account(
+        2,
+        writable,
+        name = "vault",
+        desc = "Vault token account"
+    )]
+    #[account(
+        3,
+        writable,
+        name = "destination_token",
```
joncinque: nit: can this be called `destination` or `destination_token_account`?

program/src/instruction.rs
```diff
@@ -1,184 +1,420 @@
...
     /// Only one deactivation may be in-flight at once, so if this is called
     /// with an active deactivation, it will succeed, but reset the amount and
     /// timestamp.
-    ///
-    /// 0. `[w]` Validator stake account
-    /// 1. `[s]` Authority on validator stake account
-    ///
-    /// Instruction data: amount of tokens to deactivate, as a little-endian u64
+    /// 
+    /// Instruction data: amount of tokens to deactivate, as a little-endian `u64``.
```
joncinque: micro-nit typo
```suggestion
    /// Instruction data: amount of tokens to deactivate, as a little-endian `u64`.
```

program/src/instruction.rs
```diff
@@ -1,184 +1,420 @@
...
+    #[account(
+        2,
+        writable,
+        name = "source_token",
+        desc = "Token account"
+    )]
+    #[account(
+        3,
+        signer,
+        name = "token_authority",
```
joncinque: nit: can this be `token_account_authority`? Back in 2020, I tried to call this the `transfer_authority`, but it never caught on :sweat_smile:

program/src/instruction.rs
```diff
@@ -1,184 +1,420 @@
...
+    #[account(
+        1,
+        writable,
+        name = "stake",
+        desc = "Validator stake account"
+    )]
+    #[account(
+        2,
+        writable,
+        name = "source_token",
```
joncinque: nit: can this be `source` or `source_token_account`?

program/src/instruction.rs
```diff
@@ -1,184 +1,420 @@
...
+        writable,
+        name = "config",
+        desc = "Stake config account"
+    )]
+    #[account(
+        1,
+        writable,
+        name = "stake",
+        desc = "Validator stake account"
+    )]
```
joncinque: Can this include the PDA seeds?

program/src/instruction.rs
```diff
@@ -1,184 +1,420 @@
...
+    #[account(
+        2,
+        name = "validator_vote",
+        desc = "Validator vote account"
+    )]
+    #[account(
+        3,
+        name = "system_program",
+        desc = "System program account"
```
joncinque: nit: we don't typically add the "account" suffix for programs
```suggestion
        desc = "System program"
```

program/src/instruction.rs
```diff
@@ -1,184 +1,420 @@
...
+    #[account(
+        1,
+        writable,
+        name = "config",
+        desc = "Stake config account"
+    )]
+    #[account(
+        2,
+        name = "system_program",
+        desc = "System program account"
```
joncinque: nit: we don't typically add the "account" suffix for programs
```suggestion
        desc = "System program"
```

program/src/instruction.rs
```diff
@@ -1,184 +1,420 @@
...
+        desc = "Stake token mint"
+    )]
+    #[account(
+        4,
+        name = "vault",
+        desc = "Stake token vault"
+    )]
     InitializeConfig {
-        cooldown_time: UnixTimestamp,
+        cooldown_time_seconds: UnixTimestamp,
```
joncinque, line 36: Btw, if `UnixTimestamp` is annoying, feel free to just use an `i64`

joncinque (commented):
Mostly small things, looks great overall!

program/src/instruction.rs
febo: We can specify the seeds on Kinobi and have helpers autogenerated. I will update the comments to include the seeds.

program/src/instruction.rs
febo: Not, we are disabling `fmt` on this enum so we have a predictable structure for the tags. I will fix the missing spaces.

program/src/instruction.rs
febo: I used `destination_token_account` since harvest instructions use `destination` to specify the account that receives the SOL rewards.

program/src/instruction.rs
febo: Renamed to `source_token_account` for consistency with `destination_token_account`,

program/src/instruction.rs
febo: Done.

program/src/instruction.rs
febo: It would be easier to use `i64`. I also just realised that there is a bit of an inconsistency: we have `cooldown_time_seconds: UnixTimestamp` on the `InitializeConfig` and `CooldownTimeSeconds(u64)` on the `UpdateConfig`. Should both be `i64` in this case?

joncinque: Ah right... it doesn't make sense to have a negative cooldown time, so how about we just move everything to u64s?

joncinque (approved):
This is good to go on my side, the UnixTimestamp issue can get resolved separately

buffalojoec (approved):
Lgtm!

PR #9, opened 2024-06-15
Refactor account types

febo:
This PR refactor the account types to:
1. Use bytemuck
2. Annotate with `ShankAccount`
3. Add discriminators

The IDL is updated after the refactoring.

Relevant commits due to stacking of PRs:
* [Refactor account types](https://github.com/paladin-bladesmith/stake-program/commit/62cd36ea610f1d8544e2bdcd12cc0f396102e91b)
* [Update IDL](https://github.com/paladin-bladesmith/stake-program/commit/d2e56b3fcafe9ec45698cb0ac783251da4af39ec)

* [Update discriminator](https://github.com/paladin-bladesmith/stake-program/pull/9/commits/8c4f5e44291adb9df5f311323e122d8401940294)
* [Rename to match instructions](https://github.com/paladin-bladesmith/stake-program/pull/9/commits/2cd2b3b719912095f0de51ca20f354f6c36ddc9f)
* [Add private fields](https://github.com/paladin-bladesmith/stake-program/pull/9/commits/b76a09e7bb0aab0e5c1f4624ed3a8e62d1a7e34a)
* [Update accounts discriminator](https://github.com/paladin-bladesmith/stake-program/pull/9/commits/300f01ea2c5f71a039951fbb8c19706aeca75749)

Commits related to the 2nd review:
* [Remove comments](https://github.com/paladin-bladesmith/stake-program/pull/9/commits/92d0a6b204ce5268aef8ba15df1732a0f7cb7be3)
* [Add default derive](https://github.com/paladin-bladesmith/stake-program/pull/9/commits/093c3ab9f9c9a9bf954d17fe5042a59eaafcc044)


program/src/state/config.rs
```diff
@@ -0,0 +1,52 @@
+use bytemuck::{Pod, Zeroable};
+use shank::ShankAccount;
+use solana_program::pubkey::Pubkey;
+use spl_discriminator::SplDiscriminate;
+use spl_pod::optional_keys::OptionalNonZeroPubkey;
+
+/// Configuration for a staking system.
+#[repr(C)]
+#[derive(Clone, Copy, Pod, ShankAccount, SplDiscriminate, Zeroable)]
+#[discriminator_hash_input("config::account")]
```
buffalojoec: What do you think about `stake::config` or `stake::config_account`?

program/src/state/config.rs
```diff
@@ -0,0 +1,52 @@
...
+pub struct Config {
+    /// Account disciminator.
+    ///
+    /// The discriminator is equal to `ArrayDiscriminator:: UNINITIALIZED` when
+    /// the account is empty, and equal to `Config::DISCRIMINATOR` when the account
+    /// is initialized.
+    pub discriminator: [u8; 8],
+
+    /// Authority that can modify any elements in the config.
+    // TODO: does this need to be optional?
```
buffalojoec: IMO I think authority makes sense to be optional, but I'm not sure about slash authority.

program/src/state/config.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub authority: OptionalNonZeroPubkey,
+
+    /// Authority that can slash any stake.
+    // TODO: does this need to be optional?
+    pub slash_authority: OptionalNonZeroPubkey,
+
+    /// Token account storing all stake.
+    pub vault_token: Pubkey,
```
buffalojoec: I would maybe rename this to something else so consumers understand this is a vault _token account_ and not a mint.

program/src/state/config.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    /// Running total of all stake rewards distributed.
+    pub total_stake_rewards: u64,
+
+    /// The maximum proportion that can be deactivated at once, given as basis
+    /// points (1 / 10,000).
+    pub max_deactivation_basis_points: u16,
+
+    /// Padding for alignment.
+    pub _padding: [u8; 6],
```
buffalojoec: This can be private.

program/src/state/config.rs
```diff
@@ -0,0 +1,52 @@
...
+#[repr(C)]
+#[derive(Clone, Copy, Pod, ShankAccount, SplDiscriminate, Zeroable)]
+#[discriminator_hash_input("config::account")]
+pub struct Config {
+    /// Account disciminator.
+    ///
+    /// The discriminator is equal to `ArrayDiscriminator:: UNINITIALIZED` when
+    /// the account is empty, and equal to `Config::DISCRIMINATOR` when the account
+    /// is initialized.
+    pub discriminator: [u8; 8],
```
buffalojoec: This can maybe also be private, but up to you. I can't see a reason why someone would need to access the discriminator after the account has already been deserialized/unpacked.

program/src/state/stake.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    /// Amount of staked tokens currently active
+    pub amount: u64,
+
+    /// Timestamp for when deactivation began. Used to judge if a given stake
+    /// is inactive.
+    /// NOTE: this is a special type where all zeros means `None` to avoid
+    /// wasting space, just like `Option<NonZeroU64>`
+    // TODO: Nullable trait + PodOption?
+    pub deactivation_timestamp: UnixTimestamp,
```
buffalojoec, line 26: 👍

program/src/state/config.rs
```diff
@@ -0,0 +1,52 @@
+use bytemuck::{Pod, Zeroable};
+use shank::ShankAccount;
+use solana_program::pubkey::Pubkey;
+use spl_discriminator::SplDiscriminate;
+use spl_pod::optional_keys::OptionalNonZeroPubkey;
```
buffalojoec, line 5: micro-nit: I'm not sure if I've missed this before, but we usually group these together. You can set up `rustfmt` using:

```
group_imports = "One"
imports_granularity = "One"
```

buffalojoec (commented):
Cool, thanks! Just some tiny nits!

program/src/state/config.rs
febo: Yeah, I need to set up the `nightly` formatter.

program/src/state/config.rs
febo: I think `stake::config` looks good.

program/src/state/config.rs
febo: It was renamed to `vault` on the instructions, so I used the same name here.

program/src/state/stake.rs
```diff
@@ -0,0 +1,51 @@
+use bytemuck::{Pod, Zeroable};
+use shank::ShankAccount;
+use solana_program::{clock::UnixTimestamp, pubkey::Pubkey};
+use spl_discriminator::SplDiscriminate;
+
+/// Data for an amount of tokens staked with a validator
+#[repr(C)]
+#[derive(Clone, Copy, Pod, ShankAccount, SplDiscriminate, Zeroable)]
+#[discriminator_hash_input("stake::account")]
```
buffalojoec: Ahh my bad, I forgot about this discriminator. What should we name this one, since we went with `stake::config` in the other? `stake::stake`?

febo: Is `stake::account` confusing? So we would have a `config` and `account` types.

febo: We could rename the `Stake` struct to `Account` as well.

buffalojoec: In my opinion the struct name and the discriminator should match, and the pattern should be the same for both account types. So yeah I'd say `stake::account` is confusing with the current pattern.

febo: Agree, should we name the struct `Account` or should we change the discriminator to `stake::stake`. I have a slight preference for having the struct named `Account`, but happy either way.

buffalojoec: What about just doing `stake::state::config` and `stake::state::stake`? We could also do `stake::account::config` and `stake::account::stake`. What do you think?

febo: Oh, nice idea! I like the `stake::state::*` pattern.

program/src/state/config.rs
febo: @joncinque Should we keep them as optional? Config authority being optional allows us to "freeze" the configuration when set to `None`; not sure we want to "freeze" the slash authority, unless there is a scenario where the config authority can "temporarily" disable slashing by setting the authority to `None`.

joncinque: If we want to reuse this program a simple lockup, we definitely don't want slashing enabled, so I lean toward keeping it optional

program/src/state/config.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    /// Running total of all stake rewards distributed.
+    pub total_stake_rewards: u64,
+
+    /// The maximum proportion that can be deactivated at once, given as basis
+    /// points (1 / 10,000).
+    pub max_deactivation_basis_points: u16,
+
+    /// Padding for alignment.
+    _padding: [u8; 6],
```
joncinque, line 45: If this field is going to be private, the struct needs to have a constructor or derive `Default`, otherwise it's impossible for external users to instantiate the struct. Deriving default is typically the easiest option

program/src/state/config.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+impl Config {
+    pub const LEN: usize = std::mem::size_of::<Config>();
```
joncinque, line 49: What is this used for? Is it something for shank?

joncinque (commented):
Just a couple of little points, looks good!

program/src/state/config.rs
febo: I was planning in using it the in processor to validate that the account has been initialized with the correct size. I can either removed them or add it on the PR that actually uses them?

joncinque: ah gotcha, no worries, it can stay, I was just curious, thanks!

joncinque (approved):
Looks good to me!

buffalojoec (approved):
Ship it! 🚢

PR #11, opened 2024-06-19
Add client generation

febo:
### Problem

The Kinobi config and generated code are currently from the `create-solana-program` template.

### Solution

This PR adds the initial client generation – this includes:
* Improving consistency of timestamp types so the types are correct on the generated clients
* Initial Kinobi configuration
* "Hooked" types to support optional `Pubkey`

> [!NOTE]
> All files created on the commit [Update generated clients](https://github.com/paladin-bladesmith/stake-program/pull/11/commits/8b25cc6abb0ccef6883749950c303d5bc7b85c76) have been auto-generated.
>
> CI is now running, but it is the "default" one (no `nightly fmt` and `cargo audit`). I will update it on a separate PR.


program/src/state/config.rs
```diff
@@ -27,7 +27,7 @@ pub struct Config {
 
     /// After a deactivation, defines the number of seconds that must pass before
     /// the stake is inactive and able to be withdrawn.
-    pub cooldown_time_seconds: u64,
+    pub cooldown_time_seconds: UnixTimestamp,
```
buffalojoec, line 30: Are we going with `u64` everywhere? Or do you still want `i64` in state?

clients/js/src/hooked/nullableAddress.ts
buffalojoec, line 1: I wonder if this codec makes sense to just add to https://github.com/solana-labs/solana-web3.js/tree/master/packages/codecs

clients/rust/src/hooked/nullable_address.rs
buffalojoec, line 1: On the Rust side, can't we just reuse `OptionalNonZeroPubkey` from `spl-pod`?

clients/rust/tests/create.rs
buffalojoec, line 1: NBD, but I think `cargo test` and/or `cargo test-sbf` will pass without error if there are no tests.

scripts/generate-clients.mjs
```diff
@@ -7,66 +7,95 @@ import { renderVisitor as renderRustVisitor } from "@kinobi-so/renderers-rust";
...
+                k.variablePdaSeedNode(
+                  "authority",
+                  k.publicKeyTypeNode(),
+                  "Config authority"
+                ),
+              ],
+            }),
+          ],
+        };
+      },
```
buffalojoec, line 45: I know Kinobi isn't fully documented yet, so it would be great to have some context here on what this is doing.

From what I can gather, it seems like it's declaring the `vault` field as a PDA, and the seeds are the string literal `"token-vault"` and the public key in the `authority` field?

I'm not sure on the other bits like `select: "[programNode]stake",`...

scripts/generate-clients.mjs
```diff
@@ -7,66 +7,95 @@ import { renderVisitor as renderRustVisitor } from "@kinobi-so/renderers-rust";
...
+          k.isNode(node.type, "definedTypeLinkNode") &&
+          node.type.name === "optionalNonZeroPubkey"
+        );
+      },
+      transform: (node) => {
+        k.assertIsNode(node, "structFieldTypeNode");
+        return {
+          ...node,
+          type: k.definedTypeLinkNode("nullableAddress", "hooked"),
+        };
```
buffalojoec: Same here on the Kinobi context. I'm not really sure what the `isNode` are doing in the conditional. The rest appears like it's telling Kinobi to map `OptionalNonZeroPubkey` to the nullable address type/codec you've defined in the hooks?

scripts/generate-clients.mjs
```diff
@@ -7,66 +7,95 @@ import { renderVisitor as renderRustVisitor } from "@kinobi-so/renderers-rust";
...
+        );
       },
-      arguments: {
-        amount: { defaultValue: k.noneValueNode() },
+      transform: (node) => {
+        k.assertIsNode(node, "structFieldTypeNode");
+        return {
+          ...node,
+          type: k.numberTypeNode("i64"),
+        };
```
buffalojoec: Similar comment here, but this one seems a bit more straightforward, minus the `isNode` checks.

buffalojoec (commented):
Looking good! Just some small comments and a few clarifying questions about the Kinobi configurations.

program/src/state/config.rs
febo: I left it as `i64` on state since it is quite handy to do the arithmetic without having to convert `u64` <-> `i64` when dealing with the Clock timestamp. So the "user" input is always an `u64` – they don't need to specify a negative value.

clients/js/src/hooked/nullableAddress.ts
febo: Yeah, we should definitely add it to a "shared" package. Perhaps this one is more suitable: https://github.com/solana-labs/solana-web3.js/tree/master/packages/options

I think `codecs` are more low-level de-/serialization things – I will double check with Loris.

program/src/state/config.rs
buffalojoec: Ok sounds good I'm indifferent on this. If you'd rather `i64` in state, good with me!

clients/rust/src/hooked/nullable_address.rs
febo: Oh, yeah, it might work. I will double check but from a first look they seem compatible.

joncinque (commented):
R+ with Joe's comments

clients/js/src/hooked/nullableAddress.ts
febo: Spoke to Loris, there will be a type that we can use in a future web3js release but for now he recommended to keep this one.

clients/rust/src/hooked/nullable_address.rs
febo: It could work, but needs some tweaks. The issue is with the trait `derives` – e.g., `OptionalNonZeroPubkey` does not derive `Eq`; it also uses a specific version of Borsh (`> 1.0`) while the clients are flexible to have lower requirements.

Perhaps the alternative is to add the type to `Kaigan`, which is the Rust "companion" for Kinobi.

clients/rust/tests/create.rs
febo: Yeah, it does. I will remove this placeholder.

scripts/generate-clients.mjs
febo: This is declaring what usually is referred to as a "ghost" PDA, but I think a better name is a "signer" PDA. Kinobi has the option to configure PDAs without necessarily being an account on the program.

Everything in Kinobi is a node, so we need to select the program node, then access its `pdas` property (all program nodes have a `pdas` property), then injected a new PDA.

In this case, we are saying that the program `stake`, selected with the `select: "[programNode]stake"` expression, has a PDA named `vault`; the seeds are as you described, but the `authority` is not a field but a parameter. This will generate a helper `findVaultPda` that will take an "authority" address as a seed – the constant `"token-vault"` is hard-coded so there is no need to pass it as a parameter.

PS: The helper so far only exists on the JS side but I am about to add it to the Rust client as well.

scripts/generate-clients.mjs
febo: Yep, exactly. In this case the selector is a bit more complex but its purpose is to select nodes with the specified `"authority"` or `"slashAuthority"` names; they need to be a `structFieldTypeNode` (so even if the name of the node matches a node that it not a field of a struct, it won't be selected); they need to be a "defined type" node, which means they need to be representing a type and the type name must match `"optionalNonZeroPubkey"`.

The `bottomUpTransformerVisitor` traverses the tree and you can specify a `select` expression to select nodes and apply a transformation.

In this case the transformation is to map the nodes with type `optionalNonZeroPubkey` to the `nullableAddress` that we are providing.

febo: PS: the brackets on the `["structFieldTypeNode"]` is not necessary since we are only specifying a single value – I will remove them.

scripts/generate-clients.mjs
febo: Similar to the previous comment. The `isNode` checks that the node passed as a parameter is of the given type. The "type" property of a `node` is a `node` itself.

scripts/generate-clients.mjs
buffalojoec: Cool, thanks for the explanation! Makes sense, and sounds useful the client!

If the program needs to use this PDA's derivation - say in a helper - to validate some checks in the processor, does that mean we'll have to redefine it in the program as well? Like this, for example:
https://github.com/paladin-bladesmith/rewards-program/blob/3bf1c13c2f619736c1afb759054e71e19eb55c59/program/src/state.rs#L18-L21

scripts/generate-clients.mjs
buffalojoec: Makes sense, thanks!

clients/rust/src/hooked/nullable_address.rs
buffalojoec: I'm not sold that `OptionalNonZeroPubkey` needs to be added to any core libraries (ie. `Kaigan`), but I can be swayed.

Is there no way to work around the `Eq` and Borsh issues? What about your new `PodOption<Pubkey>`? Perhaps we wait to add hooks until those types are ready?

scripts/generate-clients.mjs
febo: We can have helpers generated for the program using Shank's `seeds` macro: https://github.com/metaplex-foundation/shank/blob/master/shank-macro/src/lib.rs#L34-L71

I think we talked about this and I might have given you the answer that you couldn't, but that is for the client.

So currently, `seeds` is not going to get exported on the IDL so you need to add the PDA information twice: once with `seeds` and another on your Kinobi config.

This will change once we have Kinobi specific macros.

clients/rust/src/hooked/nullable_address.rs
febo: If we derive `Eq` on `OptionalNonZeroPubkey` should work, but it is not great to have `spl-pod` as a client dependency since it uses very specific `solana-program`, `solana-zk-token-sdk` and `borsh`. Maybe for this case it is not that relevant, but in general you would like to have a client crate more "flexible" on the dependencies.

At the moment we need a type otherwise the client generation won't work – unless we change the type of the field to a normal `Pubkey`.

buffalojoec: After discussions offline, for now it makes sense to include this new hooked type. We can always update it later if need be. I'd rather unblock the rest of the program development.

buffalojoec (approved):
This lgtm, ship it!

Just a general shout-out with the hooks and other Kinobi configurations: It doesn't seem this way, but I hope as the program processor develops, we don't have too much overhead keeping the client stuff up-to-date with the changing program.

PR #12, opened 2024-06-20
scripts: Configure and use a nightly Rust toolchain

joncinque:
#### Problem

We want to use a nightly toolchain for certain checks, like clippy and rustfmt, but rust-toolchain.toml only allows for one toolchain to be configured.

#### Solution

Add a `rust-nightly-toolchain.toml` file at the top-level, which has the same format as a rust-toolchain.toml, in case cargo every supports multiple files.

Add a function to read the toolchain.

Use it.

cc @lorisleiva for any opinions. This adds another file at the top-level, which might not be super nice, but easy for people to find and configure.

### DIFF
--- a/scripts/client/lint-rust.mjs
+++ b/scripts/client/lint-rust.mjs
@@ -1,7 +1,7 @@
 #!/usr/bin/env zx
 import 'zx/globals';
-import { workingDirectory } from '../utils.mjs';
+import { getNightlyToolchain, workingDirectory } from '../utils.mjs';
 
-// Check the client using Clippy.
+// Check the client using nightly clippy.
 cd(path.join(workingDirectory, 'clients', 'rust'));
-await $`cargo clippy ${process.argv.slice(3)}`;
+await $`cargo +${getNightlyToolchain()} clippy ${process.argv.slice(3)}`;
--- a/scripts/client/lint-rust.mjs
+++ b/scripts/client/lint-rust.mjs
@@ -2,6 +2,6 @@
 import 'zx/globals';
 import { getClippyToolchain, workingDirectory } from '../utils.mjs';
 
-// Check the client using nightly clippy.
+// Check the client using clippy.
 cd(path.join(workingDirectory, 'clients', 'rust'));
-await $`cargo +${getClippyToolchain()} clippy ${process.argv.slice(3)}`;
+await $`cargo ${getClippyToolchain()} clippy ${process.argv.slice(3)}`;
--- a/scripts/client/lint-rust.mjs
+++ b/scripts/client/lint-rust.mjs
@@ -1,7 +1,7 @@
 #!/usr/bin/env zx
 import 'zx/globals';
-import { getNightlyToolchain, workingDirectory } from '../utils.mjs';
+import { getClippyToolchain, workingDirectory } from '../utils.mjs';
 
 // Check the client using nightly clippy.
 cd(path.join(workingDirectory, 'clients', 'rust'));
-await $`cargo +${getNightlyToolchain()} clippy ${process.argv.slice(3)}`;
+await $`cargo +${getClippyToolchain()} clippy ${process.argv.slice(3)}`;
--- a/scripts/program/format.mjs
+++ b/scripts/program/format.mjs
@@ -9,5 +9,5 @@ import {
 // Format the programs.
 for (const folder of getProgramFolders()) {
   cd(`${path.join(workingDirectory, folder)}`);
-  await $`cargo +${getRustfmtToolchain()} fmt ${process.argv.slice(3)}`;
+  await $`cargo ${getRustfmtToolchain()} fmt ${process.argv.slice(3)}`;
 }
--- a/scripts/program/format.mjs
+++ b/scripts/program/format.mjs
@@ -1,9 +1,13 @@
 #!/usr/bin/env zx
 import 'zx/globals';
-import { workingDirectory, getProgramFolders } from '../utils.mjs';
+import {
+  workingDirectory,
+  getNightlyToolchain,
+  getProgramFolders,
+} from '../utils.mjs';
 
 // Format the programs.
 for (const folder of getProgramFolders()) {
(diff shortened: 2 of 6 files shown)


buffalojoec (commented):
I like this approach. I don't feel super strongly on this, but if we're going to go the `parseToml` route, we could instead include this information in `Cargo.toml` under an unregistered key, or nested under `metadata`. Something like:

```
[metadata.ci]
fmt-toolchain = "nightly-2023-10-05"
clippy-toolchain = "nightly-2023-10-05"
```

febo:
> I like this approach. I don't feel super strongly on this, but if we're going to go the `parseToml` route, we could instead include this information in `Cargo.toml` under an unregistered key, or nested under `metadata`. Something like:
> 
> ```
> [metadata.ci]
> fmt-toolchain = "nightly-2023-10-05"
> clippy-toolchain = "nightly-2023-10-05"
> ```

This is how I imaged it as well. We can have the configuration on the `Cargo.toml` and the scripts can fetch it from there.

We could also have these as parameters on the template generation, so you can specify them when creating the project (defaults to not having them and using the `stable` toolchain).

lorisleiva:
+1 on the `Cargo.toml`. Just a few things to consider:

- That information would now live under the `program` folder meaning that, if we had a multi-program repo (which can be achieved with the template), each program would be responsible for its own toolchain settings. This is probably a good thing, right? Alternatively, we can have a new metadata section on the workspace Cargo.toml if that makes more sense.
- I'm not too sure on calling the section `metadata.ci` since, technically, you can use these scripts outside of CI as well. I think it might be best to just put that information under the existing `package.metadata.solana` section.
- Also note that these metadata attributes can be hidden by default when generating the template (in case the option is less likely to be tweaked by the end-user to avoid overwhelming their Cargo.toml).

joncinque:
Thanks for the feedback! I've moved everything under `[workspace.metadata]` and removed the nightly toolchain file -- let me know what you think

buffalojoec (approved):
Wfm!

scripts/client/lint-rust.mjs
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env zx
 import 'zx/globals';
-import { workingDirectory } from '../utils.mjs';
+import { getClippyToolchain, workingDirectory } from '../utils.mjs';
 
-// Check the client using Clippy.
+// Check the client using nightly clippy.
 cd(path.join(workingDirectory, 'clients', 'rust'));
-await $`cargo clippy ${process.argv.slice(3)}`;
+await $`cargo +${getClippyToolchain()} clippy ${process.argv.slice(3)}`;
```
febo: Should this check that there was a toolchain specified otherwise default to `stable`?

I am thinking in the context of porting this change to the template.

scripts/program/format.mjs
```diff
@@ -1,9 +1,13 @@
...
+  workingDirectory,
+  getRustfmtToolchain,
+  getProgramFolders,
+} from '../utils.mjs';
 
 // Format the programs.
 for (const folder of getProgramFolders()) {
   cd(`${path.join(workingDirectory, folder)}`);
-  await $`cargo fmt ${process.argv.slice(3)}`;
+  await $`cargo +${getRustfmtToolchain()} fmt ${process.argv.slice(3)}`;
```
febo: Same as previous comment.

scripts/client/lint-rust.mjs
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env zx
 import 'zx/globals';
-import { workingDirectory } from '../utils.mjs';
+import { getClippyToolchain, workingDirectory } from '../utils.mjs';
 
-// Check the client using Clippy.
+// Check the client using nightly clippy.
```
febo: Maybe the comment should something like say "using the specified toolchain (or `stable` by default) clippy".

scripts/program/lint.mjs
```diff
@@ -1,9 +1,13 @@
...
+  workingDirectory,
+  getClippyToolchain,
+  getProgramFolders,
+} from '../utils.mjs';
 
 // Lint the programs using clippy.
 for (const folder of getProgramFolders()) {
   cd(`${path.join(workingDirectory, folder)}`);
-  await $`cargo clippy ${process.argv.slice(3)}`;
+  await $`cargo +${getClippyToolchain()} clippy ${process.argv.slice(3)}`;
```
febo: Same here.

scripts/client/lint-rust.mjs
joncinque: Oh good point. Even better, we can have this default to the empty string so it picks up the system default is rather than `stable` -- what do you think?

scripts/client/lint-rust.mjs
joncinque: I just decided to remove the `nightly` word entirely

joncinque:
I did this by falling back to the empty string, which would pick up whatever default is set based on https://rust-lang.github.io/rustup/overrides.html

scripts/client/lint-rust.mjs
febo: Perfect!

febo (approved):

scripts/client/lint-rust.mjs
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env zx
 import 'zx/globals';
-import { workingDirectory } from '../utils.mjs';
+import { getClippyToolchain, workingDirectory } from '../utils.mjs';
 
-// Check the client using Clippy.
+// Check the client using clippy.
 cd(path.join(workingDirectory, 'clients', 'rust'));
-await $`cargo clippy ${process.argv.slice(3)}`;
+await $`cargo ${getClippyToolchain()} clippy ${process.argv.slice(3)}`;
```
buffalojoec, line 7: Are you missing a `+` here?

(Same with the other two call sites)

buffalojoec: Ahh nevermind I found it. 😁

PR #13, opened 2024-06-26
Install required toolchain for lint

febo:
This PR adds a step to install the required toolchain to run `cargo fmt` and `cargo clippy` based on the `Cargo.toml` configuration. It also runs the `lint` on the program source. All linting steps moved to the start of the workflow so it fails "early" when there are issues:

image

Lint is now composed of `cargo fmt --check` and `cargo clippy` and runs both on the program and client sources.

**Note:** This is mainly a copied over from the Rewards repository, although the name of the variables are more "generic" since the toolchain might not be necessarily the "nightly" one.


.github/workflows/main.yml
```diff
@@ -192,5 +192,20 @@ jobs:
...
+        with:
+          toolchain: ${{ env.RUSTFMT_TOOLCHAIN }}
+          components: rustfmt
+      - name: Install Clippy
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.CLIPPY_TOOLCHAIN }}
+          components: clippy
+      - name: Programs
+        run: pnpm programs:lint
```
joncinque: Can this be done in a separate step at the start rather than this late? It would stink to get close to the end and fail for a formatting error

scripts/client/lint-rust.mjs
```diff
@@ -1,7 +1,13 @@
...
+  workingDirectory,
+} from "../utils.mjs";
 
-// Check the client using clippy.
-cd(path.join(workingDirectory, 'clients', 'rust'));
-await $`cargo ${getClippyToolchain()} clippy ${process.argv.slice(3)}`;
+const toolchain = getToolchainArg(getClippyToolchain());
+// Check the client using rust fmt and clippy.
+cd(path.join(workingDirectory, "clients", "rust"));
+await $`cargo ${toolchain} fmt --check ${process.argv.slice(3)}`;
```
joncinque: To be honest, I prefer keeping these two steps separate. Maybe it's just me, but I run clippy all the time while developing, but not necessarily fmt. This also makes it hard to actually format your code.

With `format.mjs`, you can easily run `format.mjs` on its own to format the code, and `format.mjs --check` to check it in CI

scripts/client/lint-rust.mjs
```diff
@@ -1,7 +1,13 @@
 #!/usr/bin/env zx
-import 'zx/globals';
-import { getClippyToolchain, workingDirectory } from '../utils.mjs';
+import "zx/globals";
```
joncinque: nit: most of the files use single quotes -- is there a reason to change these?

febo: Something to do with my VSC prettier configuration most likely. I manually fixed them.

scripts/client/lint-rust.mjs
febo: Good point. I was following the JS pattern of having only "lint", but I agree that it would be better to have them separate. SO now we have:
* `pnpm programs:format`
* `pnpm programs:clippy`
* `pnpm programs:lint` (which runs both)

Same variations for `clients:rust:*`.

buffalojoec (commented):
R+ with Jon's comments. Also, since I just landed https://github.com/paladin-bladesmith/rewards-program/pull/10 - which includes these scripts and CI setup - do you mind aligning our two layouts? I'm happy to change whatever needs to be adjusted on mine, if need be.

.github/workflows/main.yml
```diff
@@ -192,5 +192,24 @@ jobs:
...
+          components: rustfmt
+      - name: Install Clippy
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: ${{ env.CLIPPY_TOOLCHAIN }}
+          components: clippy
+      - name: Programs
+        run: |
+          pnpm programs:format --check
+          pnpm programs:clippy
```
joncinque: Sorry, maybe my comment wasn't clear, but can you move the program format / clippy checks to earlier in the pipeline? It could be something like:

```
lint_programs:
    name: Lint Programs
    runs-on: ubuntu-latest
    steps:
      - name: Git checkout
        uses: actions/checkout@v4
      - name: Setup environment
        uses: ./.github/actions/setup
        with:
          node: ${{ env.NODE_VERSION }}
      - name: Set environment variables
        shell: bash
        run: pnpm zx ./scripts/ci/set-variables.mjs
      - name: Install Rustfmt
        uses: dtolnay/rust-toolchain@master
        with:
          toolchain: ${{ env.RUSTFMT_TOOLCHAIN }}
          components: rustfmt
      - name: Install Clippy
        uses: dtolnay/rust-toolchain@master
        with:
          toolchain: ${{ env.CLIPPY_TOOLCHAIN }}
          components: clippy
      - name: Programs
        run: |
          pnpm programs:format --check
          pnpm programs:clippy
```

Even better, you could split them into two separate steps:

```
fmt_programs:
    name: Format Programs
    runs-on: ubuntu-latest
    steps:
      - name: Git checkout
        uses: actions/checkout@v4
      - name: Setup environment
        uses: ./.github/actions/setup
        with:
          node: ${{ env.NODE_VERSION }}
      - name: Set environment variables
        shell: bash
        run: pnpm zx ./scripts/ci/set-variables.mjs
      - name: Install Rustfmt
        uses: dtolnay/rust-toolchain@master
        with:
          toolchain: ${{ env.RUSTFMT_TOOLCHAIN }}
          components: rustfmt
      - name: Programs
        run: |
          pnpm programs:format --check
clippy_programs:
    name: Clippy Programs
    runs-on: ubuntu-latest
    steps:
      - name: Git checkout
        uses: actions/checkout@v4
      - name: Setup environment
        uses: ./.github/actions/setup
        with:
          node: ${{ env.NODE_VERSION }}
      - name: Set environment variables
        shell: bash
        run: pnpm zx ./scripts/ci/set-variables.mjs
      - name: Install Clippy
        uses: dtolnay/rust-toolchain@master
        with:
          toolchain: ${{ env.CLIPPY_TOOLCHAIN }}
          components: clippy
      - name: Programs
        run: pnpm programs:clippy
```

febo: I am working on moving all linting steps to the beginning, including JS, programs and rust client.

febo:
> R+ with Jon's comments. Also, since I just landed [paladin-bladesmith/rewards-program#10](https://github.com/paladin-bladesmith/rewards-program/pull/10) - which includes these scripts and CI setup - do you mind aligning our two layouts? I'm happy to change whatever needs to be adjusted on mine, if need be.

Definitely. I copied over your script + the `scripts/ci` folder. There are tweaks on the CI. I noticed that we were not running lint on the program source.

.github/workflows/main.yml
```diff
@@ -17,9 +17,69 @@ env:
...
+        uses: ./.github/actions/setup
+        with:
+          node: ${{ env.NODE_VERSION }}
+      - name: Lint JS client
+        run: pnpm clients:js:lint
+
   build_programs:
     name: Build programs
     runs-on: ubuntu-latest
+    needs: [rust_fmt_check, rust_clippy_check, lint_js]
```
joncinque, line 82: We can keep merge this as is, but keep an eye to see if it makes sense -- clippy can be a bit slow, so we might want to remove this line later

joncinque (approved):
Looks great!

.github/workflows/main.yml
joncinque: although it only took a minute on the most recent run, which should be fine :sweat_smile:

PR #14, opened 2024-06-27
Processor: `InitializeConfig`

febo:
This PR adds the processor for the `InitializeConfig` instruction.

The instruction creates the Stake config account which controls staking parameters.

> [!NOTE]


program/src/state/config.rs
```diff
@@ -14,7 +14,7 @@ pub struct Config {
     /// The discriminator is equal to `ArrayDiscriminator:: UNINITIALIZED` when
     /// the account is empty, and equal to `Config::DISCRIMINATOR` when the account
     /// is initialized.
-    discriminator: [u8; 8],
+    pub(crate) discriminator: [u8; 8],
```
joncinque: Is there reason to not make this and the padding just full `pub`?

program/src/processor/initialize_config.rs
```diff
@@ -1,21 +1,137 @@
...
+    // 1. mint
+    // - owner must the spl token 2022
+    // - must be initialized
+    // - have rewards transfer hook
+
+    require!(
+        ctx.accounts.mint.owner == &spl_token_2022::ID,
+        ProgramError::InvalidAccountOwner,
+        "mint"
+    );
```
joncinque: I think we can get rid of this requirement -- we're not requiring it on the rewards program, and we shouldn't need it here

program/src/processor/initialize_config.rs
```diff
@@ -1,21 +1,137 @@
...
+    // - amount equal to 0
+
+    let vault_data = ctx.accounts.vault.try_borrow_data()?;
+    // unpack checks if the token is initialized
+    let vault = PodStateWithExtensions::<PodAccount>::unpack(&vault_data)?;
+
+    let (vault_signer, bump) = Pubkey::find_program_address(
+        &["token-owner".as_bytes(), ctx.accounts.config.key.as_ref()],
+        program_id,
+    );
```
joncinque: nit: can this be refactored into a helper?

program/src/processor/initialize_config.rs
```diff
@@ -1,21 +1,137 @@
...
+
+    require!(
+        &vault.base.mint == ctx.accounts.mint.key,
+        StakeError::InvalidMint,
+        "vault"
+    );
+
+    let amount: u64 = vault.base.amount.into();
+    require!(amount == 0, StakeError::AmountGreaterThanZero, "vault");
```
joncinque: For this account, can you also make sure that there's no close authority, no delegate, and that the only possible extensions are "immutable owner" and "transfer hook account"? Anything else should be rejected

program/src/processor/initialize_config.rs
```diff
@@ -1,21 +1,137 @@
...
+        "config"
+    );
+
+    let config = bytemuck::from_bytes_mut::<Config>(&mut data);
+
+    require!(
+        !config.is_initialized(),
+        ProgramError::AccountAlreadyInitialized,
+        "config"
+    );
```
joncinque: This is currently fine, but can be dangerous -- if I pass in a different initialized account type with the same length as a config account, the account's discriminator will be set to something else, and pass this check.

Typically, you want to make sure that the discriminator is uninitialized (ie all 0s) to be sure that *nothing* has been initialized here.

program/src/state/config.rs
```diff
@@ -42,9 +42,24 @@ pub struct Config {
...
+
+    #[inline(always)]
+    pub fn get_signer_bump(&self) -> u8 {
+        self._padding[0]
+    }
+
+    #[inline(always)]
+    pub fn set_signer_bump(&mut self, bump: u8) {
+        self._padding[0] = bump;
+    }
```
joncinque: nit: Is there a reason to write the bump in the padding rather than just have a separate field on the struct?

clients/rust/tests/setup/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+                false,
+                "Expected custom instruction error but got '{:#?}'",
+                err
+            ),
+        };
+    };
+}
+
+#[macro_export]
+macro_rules! get_account {
```
joncinque, line 48: nit: maybe I'm missing a future use case, but why is this a macro instead of a function?

clients/rust/tests/setup/token.rs
```diff
@@ -0,0 +1,147 @@
...
+        &instructions,
+        Some(&context.payer.pubkey()),
+        &[&context.payer, mint],
+        context.last_blockhash,
+    );
+
+    context.banks_client.process_transaction(tx).await
+}
+
+pub async fn create_token(
```
joncinque: nit: can you call this `create_token_account`? I know the nomenclature stinks, but otherwise it's unclear if this is creating a mint or an account

clients/rust/tests/initialize_config.rs
```diff
@@ -0,0 +1,581 @@
...
+use setup::{create_mint, create_token, mint_to, MINT_EXTENSIONS, TOKEN_ACCOUNT_EXTENSIONS};
+use solana_program_test::{tokio, ProgramTest};
+use solana_sdk::{
+    instruction::InstructionError,
+    signature::{Keypair, Signer},
+    system_instruction,
+    transaction::Transaction,
+};
+
+mod initialize_config {
```
joncinque: micro-nit: why is this wrapped in a `mod`?

clients/rust/tests/initialize_config.rs
```diff
@@ -0,0 +1,581 @@
...
+            &mint,
+            &authority,
+            Some(&authority),
+            0,
+            MINT_EXTENSIONS,
+        )
+        .await
+        .unwrap();
+
+        let token = Keypair::new();
```
joncinque: nit: can this be `token_account` to differentiate better from `mint`?

clients/rust/tests/initialize_config.rs
```diff
@@ -0,0 +1,581 @@
...
+            .config(config.pubkey())
+            .config_authority(authority)
+            .slash_authority(authority)
+            .mint(mint.pubkey())
+            .vault(token.pubkey())
+            .cooldown_time_seconds(1) // 1 second
+            .max_deactivation_basis_points(500) // 5%
+            .instruction();
+
+        // When we try to initialize the config with the wrong token authority.
```
joncinque: micro-nit: Should probably say "When we try to initialize the config with an uninitialized mint.", right?

clients/rust/tests/initialize_config.rs
```diff
@@ -0,0 +1,581 @@
...
+            .config(config.pubkey())
+            .config_authority(authority_pubkey)
+            .slash_authority(authority_pubkey)
+            .mint(mint.pubkey())
+            .vault(token.pubkey())
+            .cooldown_time_seconds(1) // 1 second
+            .max_deactivation_basis_points(500) // 5%
+            .instruction();
+
+        // When we try to initialize the config with a non-empty token.
```
joncinque: micro-nit: Should probably say, "When we try to initialize the config with an incorrectly-sized account."

joncinque (commented):
Looking great overall! Mostly small stuff

program/src/state/config.rs
febo: Currently we weren't unsure whether this type would be used directly by "client" code or not. Since it is not, I made it full `pub`.

program/src/state/config.rs
febo: Oh, I didn't realised that I could have both `u16` and `u8` as the last 2 fields of the account. I will move it to a "normal" field.

clients/rust/tests/setup/mod.rs
febo: It was just for convenience so you don't need to do `get_account(...).await` but it can definitely be a function if that makes more sense.

clients/rust/tests/initialize_config.rs
febo: For convenience, so you can run all initialize config tests by using:
* `--test initialize_config`

And any specific one by using:
* `--test initialize_config -- fail_initialize_config_with_initialized_account`

I can remove the `mod` if that makes more sense – maybe there is a better way to achieve that without the `mod`?

clients/rust/tests/initialize_config.rs
febo: Yep, copy-paste left over.

clients/rust/tests/initialize_config.rs
febo: Yep!

program/src/processor/initialize_config.rs
febo: Good point! I added an `is_uninitialized()` helper to the `Config` account that compares against an empty discriminator and used it here.

clients/rust/tests/initialize_config.rs
joncinque: Maybe I'm missing something, but the `mod` declaration doesn't handle that, being in a separate file does that.

For example, if I have `tests/mytest.rs`, I run all the tests in that file with:

```
cargo test --test mytest
```

And one test with

```
cargo test --test mytest -- a_test
```

If I add a mod inside like you have here, then the nested test is fully specified as:

```
cargo test --test mytest -- mytest::a_nested_test
```

clients/rust/tests/setup/mod.rs
joncinque: Ah gotcha, it's all good either way!

clients/rust/tests/initialize_config.rs
febo: Oh, I have been adding a `mod` for no reason then. 😅

I could run nested tests without fully-qualifying them, like:
```bash
cargo test --test mytest -- a_nested_test
```
But the `mod` is actually not needed to achieve this, I will remove it.

program/src/processor/initialize_config.rs
febo: Done! The code enforces that both `ImmutableOwner` and `TransferHookAccount` are present. When either is missing or any other one is found, it fails.

clients/rust/Cargo.toml
```diff
@@ -22,4 +22,5 @@ thiserror = "^1.0"
 assert_matches = "1.5.0"
 solana-program-test = "^1.16"
 solana-sdk = "^1.16"
+spl-memo = "^4.0.0"
```
joncinque: Is this needed?

program/src/processor/initialize_config.rs
```diff
@@ -65,24 +66,56 @@ pub fn process_initialize_config(
...
+
+    vault
+        .get_extension_types()?
+        .iter()
+        .try_for_each(|extension_type| {
+            if !VALID_VAULT_TOKEN_EXTENSIONS.contains(extension_type) {
+                msg!("Invalid token extension: {:?}", extension_type);
+                return Err(StakeError::InvalidTokenAccountExtension);
+            }
+            required_extensions.retain(|e| e != extension_type);
```
joncinque: Maybe I'm misunderstanding the idea here, but if the token account has both extensions, then I think this will actually remove all the `required_extensions` and fail the check later, which isn't the intention.

For example, if `get_extension_types()` returns `[TransferHookAccount, ImmutableOwner]`, then the first iteration will remove `ImmutableOwner`, and the second iteration will remove `TransferHookAccount`.

I think we can get rid of the `required_extensions` check, but keep the `VALID_VAULT_TOKEN_EXTENSIONS` check.

joncinque (commented):
Just a couple of last bits, then this is good to go!

program/src/processor/initialize_config.rs
febo: Yes, all the `required_extensions` need to be removed to indicate that all of them were found - the check is requiring `required_extensions.is_empty()` at the end.

The "loop" is retaining the extensions that are different than the current one, so if the list is `[TransferHookAccount, ImmutableOwner]`, `TransferHookAccount` will be removed first and then `ImmutableOwner`.

This is to assert that we got both ones set. It is technically not possible to have `TransferHookAccount` missing since we are validating the mint, but the token account might not have `ImmutableOwner`.

If there is no need to enforce this, we can remove the "required extensions" assert.

clients/rust/Cargo.toml
febo: Nope, left over from when I was trying to perform a transfer with the memo extension enabled. Then I went down the route of changing the owner of the account "manually", which was much easier. I will remove it.

program/src/processor/initialize_config.rs
joncinque: Hehe looks like I did totally misunderstand, sorry about that.

Either way, let's keep it simple and only allow those two extensions, immutable ownership shouldn't be required.

program/src/processor/initialize_config.rs
```diff
@@ -72,7 +72,7 @@ pub fn process_initialize_config(
     // - amount equal to 0
 
     let vault_data = ctx.accounts.vault.try_borrow_data()?;
-    // unpack checks if the token is initialized
+    // unpack checks if the token account is initialized
```
joncinque, line 75: :heart:

joncinque (approved):
Looks good!

takeoff

PR #15, opened 2024-06-27
Processor: `UpdateConfig`

febo:
This PR adds the processor for the `UpdateConfig` instruction.

The instruction update the staking parameters of the the Stake config account.

> [!NOTE]

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
 
 env:
   NODE_VERSION: 18
--- a/clients/rust/tests/update_config.rs
+++ b/clients/rust/tests/update_config.rs
@@ -1,0 +1,349 @@
+#![cfg(feature = "test-sbf")]
+
+mod setup;
+
+use paladin_stake::{
+    accounts::Config,
+    errors::StakeError,
+    instructions::{InitializeConfigBuilder, UpdateConfigBuilder},
+    pdas::find_vault_pda,
+    types::ConfigField,
+};
+use setup::{create_mint, create_token, MINT_EXTENSIONS, TOKEN_ACCOUNT_EXTENSIONS};
+use solana_program_test::{tokio, ProgramTest};
+use solana_sdk::{
+    instruction::InstructionError,
+    signature::{Keypair, Signer},
+    system_instruction,
+    transaction::Transaction,
+};
+
+mod update_config {
+
+    use super::*;
+
+    #[tokio::test]
+    async fn update_config() {
+        let mut context = ProgramTest::new("stake_program", paladin_stake::ID, None)
+            .start_with_context()
+            .await;
+
+        // Given an empty config account and a mint.
+
+        let config = Keypair::new();
+        let authority = Keypair::new();
+
+        let mint = Keypair::new();
+        create_mint(
+            &mut context,
+            &mint,
+            &authority.pubkey(),
+            Some(&authority.pubkey()),
+            0,
+            MINT_EXTENSIONS,
+        )
+        .await
+        .unwrap();
+
+        let token = Keypair::new();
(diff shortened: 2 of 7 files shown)


.github/workflows/main.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
```
joncinque, line 7: Was this an accident?

program/src/processor/update_config.rs
```diff
@@ -12,9 +19,66 @@ use crate::instruction::{
...
+            ConfigField::CooldownTimeSeconds(seconds) => {
+                config.cooldown_time_seconds = seconds as UnixTimestamp;
+            }
+            ConfigField::MaxDeactivationBasisPoints(points) => {
+                config.max_deactivation_basis_points = points;
+            }
+        }
+    } else {
+        // TODO: do we need to log a message to say that the config authority is not set
+        // and the update did not happen? Or fail the transaction?
```
joncinque: As discussed earlier, let's fail this

clients/rust/tests/update_config.rs
```diff
@@ -0,0 +1,349 @@
...
+            &authority.pubkey(),
+            Some(&authority.pubkey()),
+            0,
+            MINT_EXTENSIONS,
+        )
+        .await
+        .unwrap();
+
+        let token = Keypair::new();
+        create_token(
```
joncinque: nit: the whole token -> token_account bit should get rectified with the previous PR, but still flagging

clients/rust/tests/update_config.rs
```diff
@@ -0,0 +1,349 @@
...
+            &[create_ix, initialize_ix],
+            Some(&context.payer.pubkey()),
+            &[&context.payer, &config],
+            context.last_blockhash,
+        );
+        context.banks_client.process_transaction(tx).await.unwrap();
+
+        let account = get_account!(context, config.pubkey());
+        let config_account = Config::from_bytes(account.data.as_ref()).unwrap();
+        assert_eq!(config_account.cooldown_time_seconds, 1);
```
joncinque: nit: up to you, but this could probably get refactored into another function

clients/rust/tests/update_config.rs
```diff
@@ -0,0 +1,349 @@
...
+            &[&context.payer, &authority],
+            context.last_blockhash,
+        );
+        context.banks_client.process_transaction(tx).await.unwrap();
+
+        // Then the config was updated.
+
+        let account = get_account!(context, config.pubkey());
+        let config_account = Config::from_bytes(account.data.as_ref()).unwrap();
+        assert_eq!(config_account.cooldown_time_seconds, 10);
```
joncinque: Can you also test setting the other config field?

clients/rust/tests/update_config.rs
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    #[tokio::test]
+    async fn fail_update_config_with_wrong_authority() {
```
joncinque: This is kind of covered by this test, but can you add a case for failing when there's no authority set?

joncinque (commented):
Looks really good overall! Mostly small stuff

.github/workflows/main.yml
febo: Intentional so "stacked" PRs run CI; otherwise only PRs directed to `main` run CI. Not sure if there is a better way to achieve this.

joncinque: Ah right, yep, that makes sense, thanks

clients/rust/tests/update_config.rs
febo: Yep, good call. I will do it in a separate PR since we need to change all tests to use it.

joncinque (approved):
Looks great! 

cool

PR #16, opened 2024-06-27
Processor: `SetAuthority`

febo:
This PR adds the processor for the `SetAuthority` instruction.

The instruction sets new authority on a `Config` or `Stake` account.

Commits relevant (since the rebase was from another branch):
* [Add processor implementation](https://github.com/paladin-bladesmith/stake-program/pull/16/commits/50c590a33080b8529ea1fa3b85d3942bf9b55f18)
* [Add tests](https://github.com/paladin-bladesmith/stake-program/pull/16/commits/5568e0658fd3ccba7d910e4bfa00d8f7cb9ddcb1)
* [Address review comments](https://github.com/paladin-bladesmith/stake-program/pull/16/commits/1ffe9916796bfde9418029bbc445dbf49c707816)
* [Update error variant](https://github.com/paladin-bladesmith/stake-program/pull/16/commits/d57da2b54fe0ff7e824091663c6e31bbceff245e)
* [Add stake tests](https://github.com/paladin-bladesmith/stake-program/pull/16/commits/b8a0ea42595f598366c93833c191fad64ba2898a)

> [!NOTE]

### DIFF
--- a/clients/rust/tests/set_authority.rs
+++ b/clients/rust/tests/set_authority.rs
@@ -1,0 +1,428 @@
+#![cfg(feature = "test-sbf")]
+
+mod setup;
+
+use paladin_stake::{
+    accounts::Config,
+    errors::StakeError,
+    instructions::{InitializeConfigBuilder, SetAuthorityBuilder},
+    pdas::find_vault_pda,
+    types::AuthorityType,
+};
+use setup::{create_mint, create_token, MINT_EXTENSIONS, TOKEN_ACCOUNT_EXTENSIONS};
+use solana_program_test::{tokio, ProgramTest};
+use solana_sdk::{
+    signature::{Keypair, Signer},
+    system_instruction,
+    transaction::Transaction,
+};
+
+mod set_authority {
+
+    use solana_sdk::pubkey::Pubkey;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn set_config_authority_on_config() {
+        let mut context = ProgramTest::new("stake_program", paladin_stake::ID, None)
+            .start_with_context()
+            .await;
+
+        // Given an empty config account and a mint.
+
+        let config = Keypair::new();
+        let authority = Keypair::new();
+        let authority_pubkey = authority.pubkey();
+
+        let mint = Keypair::new();
+        create_mint(
+            &mut context,
+            &mint,
+            &authority_pubkey,
+            Some(&authority_pubkey),
+            0,
+            MINT_EXTENSIONS,
+        )
+        .await
+        .unwrap();
+
+        let token = Keypair::new();
+        create_token(
+            &mut context,
+            &find_vault_pda(&config.pubkey()).0,
+            &token,
+            &mint.pubkey(),
+            TOKEN_ACCOUNT_EXTENSIONS,
+        )
+        .await
+        .unwrap();
+
+        let create_ix = system_instruction::create_account(
+            &context.payer.pubkey(),
+            &config.pubkey(),
+            context
+                .banks_client
+                .get_rent()
+                .await
+                .unwrap()
+                .minimum_balance(Config::LEN),
+            Config::LEN as u64,
+            &paladin_stake::ID,
+        );
+
+        let initialize_ix = InitializeConfigBuilder::new()
+            .config(config.pubkey())
+            .config_authority(authority_pubkey)
+            .slash_authority(authority_pubkey)
+            .mint(mint.pubkey())
+            .vault(token.pubkey())
+            .cooldown_time_seconds(1) // 1 second
+            .max_deactivation_basis_points(500) // 5%
+            .instruction();
+
+        // And we initialize a config.
+
+        let tx = Transaction::new_signed_with_payer(
+            &[create_ix, initialize_ix],
+            Some(&context.payer.pubkey()),
+            &[&context.payer, &config],
+            context.last_blockhash,
+        );
+        context.banks_client.process_transaction(tx).await.unwrap();
+
+        let account = get_account!(context, config.pubkey());
+        let config_account = Config::from_bytes(account.data.as_ref()).unwrap();
+        assert_eq!(config_account.authority, authority_pubkey.into());
+
+        // When we set a new authority on the config.
+
+        let new_authority = Keypair::new();
+
+        let set_authority_ix = SetAuthorityBuilder::new()
+            .account(config.pubkey())
+            .authority(authority_pubkey)
+            .new_authority(new_authority.pubkey())
+            .authority_type(AuthorityType::Config)
+            .instruction();
+
+        let tx = Transaction::new_signed_with_payer(
+            &[set_authority_ix],
+            Some(&context.payer.pubkey()),
+            &[&context.payer, &authority],
+            context.last_blockhash,
+        );
+        context.banks_client.process_transaction(tx).await.unwrap();
+
+        // Then the authority is updated.
+
+        let account = get_account!(context, config.pubkey());
+        let config_account = Config::from_bytes(account.data.as_ref()).unwrap();
+        assert_eq!(config_account.authority, new_authority.pubkey().into());
+    }
+
+    #[tokio::test]
+    async fn set_slash_authority_on_config() {
+        let mut context = ProgramTest::new("stake_program", paladin_stake::ID, None)
+            .start_with_context()
+            .await;
+
+        // Given an empty config account and a mint.
+
+        let config = Keypair::new();
+        let authority = Keypair::new();
+        let authority_pubkey = authority.pubkey();
+
+        let mint = Keypair::new();
+        create_mint(
+            &mut context,
+            &mint,
+            &authority_pubkey,
+            Some(&authority_pubkey),
+            0,
+            MINT_EXTENSIONS,
+        )
+        .await
+        .unwrap();
+
+        let token = Keypair::new();
+        create_token(
+            &mut context,
+            &find_vault_pda(&config.pubkey()).0,
+            &token,
+            &mint.pubkey(),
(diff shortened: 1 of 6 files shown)


program/src/processor/set_authority.rs
```diff
@@ -13,9 +19,85 @@ use crate::instruction::{
 ///   1. `[s]` config_authority
```
joncinque: nit: I can't comment higher up but can you change line 12 to "0. `[w]` config or stake account" and then line 13 to "1. `[s]` current authority"?

program/src/processor/set_authority.rs
```diff
@@ -13,9 +19,85 @@ use crate::instruction::{
...
+            let stake = bytemuck::try_from_bytes_mut::<Stake>(data)
+                .map_err(|_error| ProgramError::InvalidAccountData)?;
+
+            require!(
+                *ctx.accounts.authority.key == stake.authority,
+                StakeError::InvalidAuthority,
+                "authority (stake)"
+            );
+
+            stake.authority = *ctx.accounts.new_authority.key;
```
joncinque: This might actually be OK as is, but can you add some form of `is_initialized()` check for the `Stake` too?

program/src/processor/set_authority.rs
```diff
@@ -13,9 +19,85 @@ use crate::instruction::{
...
+            require!(
+                config.is_initialized(),
+                ProgramError::UninitializedAccount,
+                "config"
+            );
+
+            // Asserts that the authority is set - only updates the authority jf there is
+            // one set and it matches the signing authority.
+            //
+            // TODO: do we fail if the authority is not set?
```
joncinque: Same as on #15 -- let's definitely fail

program/src/processor/set_authority.rs
```diff
@@ -13,9 +19,85 @@ use crate::instruction::{
...
+
+            let current_authority = config.authority.into();
```
joncinque: Sorry, this wasn't clear, but I was thinking this would operate similarly to the different authorities in an SPL token mint, where each authority has control over itself and that's it.

So the config authority can set another config authority, and the slash authority can set another slash authority. This way we can eventually delegate those authorities to different programs. What do you think?

clients/rust/tests/set_authority.rs
```diff
@@ -0,0 +1,428 @@
...
+
+        // When we set a new slash authority on the config.
+
+        let new_slash_authority = Keypair::new();
+
+        let set_authority_ix = SetAuthorityBuilder::new()
+            .account(config.pubkey())
+            .authority(authority_pubkey)
+            .new_authority(new_slash_authority.pubkey())
+            .authority_type(AuthorityType::Config)
```
joncinque: Shouldn't this be `AuthorityType::Slash`?

clients/rust/tests/set_authority.rs
```diff
@@ -0,0 +1,428 @@
...
+        let account = get_account!(context, config.pubkey());
+        let config_account = Config::from_bytes(account.data.as_ref()).unwrap();
+        assert_eq!(
+            config_account.authority,
+            new_slash_authority.pubkey().into()
+        );
+    }
+
+    #[tokio::test]
+    async fn failt_set_config_authority_with_wrong_authority() {
```
joncinque: tiny nit
```suggestion
    async fn fail_set_config_authority_with_wrong_authority() {
```

clients/rust/tests/set_authority.rs
```diff
@@ -0,0 +1,428 @@
...
+
+        // When we try to set a new slash authority when config authority is None.
+
+        let new_slash_authority = Keypair::new();
+
+        let set_authority_ix = SetAuthorityBuilder::new()
+            .account(config.pubkey())
+            .authority(authority_pubkey)
+            .new_authority(new_slash_authority.pubkey())
+            .authority_type(AuthorityType::Config)
```
joncinque: Shouldn't this be `AuthorityType::Slash`?

clients/rust/tests/set_authority.rs
```diff
@@ -0,0 +1,428 @@
...
+};
+
+mod set_authority {
+
+    use solana_sdk::pubkey::Pubkey;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn set_config_authority_on_config() {
```
joncinque: Can you add a test case for setting the stake authority?

joncinque (commented):
Looking good! Just a few things to clarify

program/src/processor/set_authority.rs
febo: Got it, makes sense. I will change the implementation to reflect this.

clients/rust/tests/set_authority.rs
febo: Done!

program/src/processor/set_authority.rs
```diff
@@ -13,9 +19,85 @@ use crate::instruction::{
...
-    _authority_type: AuthorityType,
+    program_id: &Pubkey,
+    ctx: Context<SetAuthorityAccounts>,
+    authority_type: AuthorityType,
 ) -> ProgramResult {
+    // Accounts validation.
+
+    // 1. authority
+    // - must be a signer
+    // - must match the authority in the account (checked in the match statement below)
```
buffalojoec: The instruction has `account` first, then `authority` - in case you wanted to make sure these comments and the ordering of checks matched.

program/src/processor/set_authority.rs
```diff
@@ -13,9 +19,85 @@ use crate::instruction::{
...
+                match authority_type {
+                    AuthorityType::Config => {
+                        config.authority = OptionalNonZeroPubkey(*ctx.accounts.new_authority.key)
+                    }
+                    AuthorityType::Slash => {
+                        config.slash_authority =
+                            OptionalNonZeroPubkey(*ctx.accounts.new_authority.key)
+                    }
+                    _ => (), /* unreachable */
+                }
```
buffalojoec: What do you think about combining this and the encompassing match blocks? Can you make this a helper or closure that gets called into?

```rust
let config = bytemuck::try_from_bytes_mut::<Config>(data)
    .map_err(|_error| ProgramError::InvalidAccountData)?;

require!(
    config.is_initialized(),
    ProgramError::UninitializedAccount,
    "config"
);
```

This way, we can match on the `AuthorityType` once, and do away with the unreachable bit.

buffalojoec: You could even roll a unified helper along these lines, up to you!

```rust
fn unpack_initialized_mut::<'a, T: Pod>(data: &'a mut [u8], name: &str) -> &'a mut T {
    let state = bytemuck::try_from_bytes_mut::<T>(data)
        .map_err(|_error| ProgramError::InvalidAccountData)?;

    require!(
        state.is_initialized(),
        ProgramError::UninitializedAccount,
        name
    );

    state
}
```

program/src/processor/set_authority.rs
```diff
@@ -13,9 +19,85 @@ use crate::instruction::{
...
+            }
+        }
+        AuthorityType::Stake => {
+            let stake = bytemuck::try_from_bytes_mut::<Stake>(data)
+                .map_err(|_error| ProgramError::InvalidAccountData)?;
+
+            require!(
+                *ctx.accounts.authority.key == stake.authority,
+                StakeError::InvalidAuthority,
```
buffalojoec, line 105: Ehh not really super important, but if you bumped this crate to 2.0, you could use `ProgramError::IncorrectAuthority`.

https://github.com/anza-xyz/agave/blob/1e9d00d324398cded266820c73d0856988995484/sdk/program/src/program_error.rs#L70

buffalojoec (commented):
Looking good, just a few small comments from me!

program/src/processor/set_authority.rs
febo: I removed the numbering to not give the impression that it is the account order.

program/src/processor/set_authority.rs
febo: I think we need a trait to do a function, we could use a macro instead.

program/src/processor/set_authority.rs
febo: Which version should we be using?

program/src/processor/set_authority.rs
buffalojoec: Thanks. Clearly it got me on both processors!

program/src/processor/set_authority.rs
buffalojoec: > I think we need a trait to do a function, we could use a macro instead.

`<'a, T: Pod>` would work for the layout, but yeah you're right we'd have trouble with `.is_initialized()`.

program/src/processor/set_authority.rs
buffalojoec: I think 1.18 is fine for now. That's what mb will be on for a while.

program/src/processor/set_authority.rs
buffalojoec: If you think it's worth it, `solana-program` has such a trait:

https://github.com/anza-xyz/agave/blob/0ff5a4ba92dc68b446113a93547cd4affc7b0602/sdk/program/src/program_pack.rs#L5-L9

buffalojoec: Ooh nice, the macro is pretty nifty! Wfm.

buffalojoec (approved):

program/src/processor/set_authority.rs
febo: Good to know about the `IsInitialized` trait. I considered adding a similar trait, but then `Config` has `is_uninitialized` and it looked odd to not have it as part of the trait. At the same time, `Stake` does not have the `is_uninitialized`. I then settle for the macro. ☺️

PR #19, opened 2024-06-27
Enable nightly formatter settings

febo:
This PR adds the nightly formatting settings.

All changes to the code are the results of running `cargo fmt` with the new settings.


rustfmt.toml
```diff
@@ -1,10 +1,5 @@
...
-# imports_indent = "Block"
-# imports_layout = "Mixed"
-# imports_granularity = "Crate"
-# group_imports = "Preserve"
-# reorder_impl_items = false
+imports_indent = "Block"
+imports_layout = "Mixed"
+imports_granularity = "Crate"
+group_imports = "Preserve"
+reorder_impl_items = false
```
joncinque: nit: It looks like `imports_granularity` is the only setting that's *not* the default for rustfmt https://rust-lang.github.io/rustfmt/

Personally, I like:

```
edition = "2021"
group_imports = "One"
imports_granularity = "One"
```

buffalojoec: I recommend these from SPL:

```toml
comment_width = 80
edition = "2021"
group_imports = "One"
imports_granularity = "One"
wrap_comments = true
```

scripts/generate-clients.mjs
```diff
@@ -114,5 +114,6 @@ kinobi.accept(
   renderRustVisitor(path.join(rustClient, "src", "generated"), {
     formatCode: true,
     crateFolder: rustClient,
+    toolchain: "+nightly-2023-10-05",
```
buffalojoec: I think we're starting to diverge here.

https://github.com/paladin-bladesmith/rewards-program/blob/8934926b08280fd6f1abb1b8f17dba761f573f54/scripts/generate-clients.mjs#L36

.github/workflows/main.yml
```diff
@@ -162,7 +162,9 @@ jobs:
       - name: Generate IDLs
         run: pnpm generate:idls
       - name: Ensure working directory is clean
-        run: test -z "$(git status --porcelain)"
+        run: |
+          git status --porcelain
+          test -z "$(git status --porcelain)"
```
buffalojoec, line 180: What exactly does duplicating this command do?

.github/workflows/main.yml
```diff
@@ -178,6 +178,14 @@ jobs:
         with:
           node: ${{ env.NODE_VERSION }}
           solana: ${{ env.SOLANA_VERSION }}
+      - name: Set environment variables
+        shell: bash
+        run: pnpm zx ./scripts/ci/set-variables.mjs
```
buffalojoec: You could add this script invocation to the `package.json` and just do `pnpm ci:set-env`.
Your call, just let me know if I need to update my repository.

https://github.com/paladin-bladesmith/rewards-program/blob/8934926b08280fd6f1abb1b8f17dba761f573f54/package.json#L4

buffalojoec (commented):
Looking good, just a few tiny questions!

scripts/generate-clients.mjs
joncinque: Can this use the `getRustfmtToolchain` helper?

joncinque (commented):
Looks good! Just a couple of little suggestions

rustfmt.toml
febo: Cool, we should update the template with a set of defaults. The ones that I am using are from the template – we could use the ones that you are recommending.

scripts/generate-clients.mjs
febo: You way is definitely better, I forgot to use the helper here.

.github/workflows/main.yml
febo: It will log the differences. It will only produce an output when `test -z "$(git status --porcelain)"` fails, so it is useful to see "why" it failed.

.github/workflows/main.yml
febo: Perhaps it would be better not to add it on `package.json` since it is not something that needs to be "expose" to users. What do you think?

rustfmt.toml
febo: Following Joe's comments, we can use the "defaults" from SPL.

scripts/generate-clients.mjs
febo: Yep, that is much better.

joncinque (approved):

clients/rust/src/generated/instructions/deactivate_stake.rs
```diff
@@ -144,7 +145,8 @@ impl DeactivateStakeBuilder {
 
 /// `deactivate_stake` CPI accounts.
 pub struct DeactivateStakeCpiAccounts<'a, 'b> {
-    /// Validator stake account (pda of `['stake::state::stake', validator, config]`)
+    /// Validator stake account (pda of `['stake::state::stake', validator,
+    /// config]`)
```
buffalojoec, line 149: Ahh, some of these are a bit rough after formatting, but this can be handled another time!

program/src/entrypoint.rs
```diff
@@ -1,14 +1,12 @@
...
-        account_info::AccountInfo,
-        entrypoint::{self, ProgramResult},
-        program_error::PrintProgramError,
+        account_info::AccountInfo, entrypoint::ProgramResult, program_error::PrintProgramError,
         pubkey::Pubkey,
     },
 };
 
-entrypoint!(process_instruction);
+solana_program::entrypoint!(process_instruction);
```
buffalojoec, line 9: Nice, yeah this is one little annoyance with the formatting rules, but glad you caught it!

buffalojoec (approved):

.github/workflows/main.yml
buffalojoec: Yeah you're probably right. I'll update mine!

febo:
This was superseded by the latests changes to the solana program template.

PR #20, opened 2024-07-02
Processor: `InitializeStake`

febo:
This PR adds the processor for the `InitializeStake` instruction.

### DIFF
--- a/clients/rust/src/pdas.rs
+++ b/clients/rust/src/pdas.rs
@@ -3,3 +3,14 @@ use solana_program::pubkey::Pubkey;
 pub fn find_vault_pda(config: &Pubkey) -> (Pubkey, u8) {
     Pubkey::find_program_address(&["token-owner".as_bytes(), config.as_ref()], &crate::ID)
 }
+
+pub fn find_stake_pda(validator: &Pubkey, config: &Pubkey) -> (Pubkey, u8) {
+    Pubkey::find_program_address(
+        &[
+            "stake::state::stake".as_bytes(),
+            validator.as_ref(),
+            config.as_ref(),
+        ],
+        &crate::ID,
+    )
+}
--- a/clients/rust/tests/initialize_stake.rs
+++ b/clients/rust/tests/initialize_stake.rs
@@ -1,0 +1,82 @@
+#![cfg(feature = "test-sbf")]
+
+mod setup;
+
+use paladin_stake::{accounts::Stake, instructions::InitializeStakeBuilder, pdas::find_stake_pda};
+use setup::config::create_config;
+use solana_program_test::{tokio, ProgramTest};
+use solana_sdk::{
+    account::Account,
+    clock::Epoch,
+    signature::{Keypair, Signer},
+    system_instruction,
+    transaction::Transaction,
+    vote::state::VoteState,
+};
+
+#[tokio::test]
+async fn initialize_config_with_mint_and_token() {
+    let mut context = ProgramTest::new("stake_program", paladin_stake::ID, None)
+        .start_with_context()
+        .await;
+
+    // Given a config account and a validator's vote account.
+
+    let config = create_config(&mut context).await;
+    let vote = Keypair::new().pubkey();
+    let validator = Keypair::new().pubkey();
+
+    let mut data = vec![0u8; VoteState::size_of()];
+    // 0-4 is the version offset
+    data[4..36].copy_from_slice(validator.as_ref()); // node_pubkey
+    data[36..68].copy_from_slice(validator.as_ref()); // authorized_withdrawer
+
+    let vote_account = Account {
+        lamports: 5_000_000_000, // 5 SOL
+        data,
+        owner: solana_sdk::vote::program::ID,
+        executable: false,
+        rent_epoch: Epoch::default(),
+    };
+
+    context.set_account(&vote, &vote_account.into());
+
+    // When we initialize the stake account.
+
+    let (stake_pda, _) = find_stake_pda(&validator, &config);
+
+    let transfer_ix = system_instruction::transfer(
+        &context.payer.pubkey(),
+        &stake_pda,
+        context
+            .banks_client
+            .get_rent()
+            .await
+            .unwrap()
+            .minimum_balance(Stake::LEN),
+    );
+
+    let initialize_ix = InitializeStakeBuilder::new()
+        .config(config)
+        .stake(stake_pda)
+        .validator_vote(vote)
+        .instruction();
+
+    let tx = Transaction::new_signed_with_payer(
+        &[transfer_ix, initialize_ix],
+        Some(&context.payer.pubkey()),
+        &[&context.payer],
+        context.last_blockhash,
+    );
+    context.banks_client.process_transaction(tx).await.unwrap();
+
+    // Then an account was created with the correct data.
+
+    let account = get_account!(context, stake_pda);
+    assert_eq!(account.data.len(), Stake::LEN);
+
+    let account_data = account.data.as_ref();
+    let stake_account = Stake::from_bytes(account_data).unwrap();
+    assert_eq!(stake_account.validator, validator);
+    assert_eq!(stake_account.authority, validator);
+}
--- a/clients/rust/tests/initialize_stake.rs
+++ b/clients/rust/tests/initialize_stake.rs
@@ -2,20 +2,26 @@
 
 mod setup;
 
-use paladin_stake::{accounts::Stake, instructions::InitializeStakeBuilder, pdas::find_stake_pda};
-use setup::config::create_config;
+use paladin_stake::{
+    accounts::{Config, Stake},
+    instructions::InitializeStakeBuilder,
+    pdas::find_stake_pda,
+};
+use setup::{
+    config::create_config,
+    vote::{create_vote_account, create_vote_account_with_program_id},
+};
 use solana_program_test::{tokio, ProgramTest};
 use solana_sdk::{
-    account::Account,
-    clock::Epoch,
+    instruction::InstructionError,
+    pubkey::Pubkey,
     signature::{Keypair, Signer},
-    system_instruction,
+    system_instruction, system_program,
     transaction::Transaction,
-    vote::state::VoteState,
 };
 
 #[tokio::test]
-async fn initialize_config_with_mint_and_token() {
+async fn initialize_stake_with_validator_vote() {
     let mut context = ProgramTest::new("stake_program", paladin_stake::ID, None)
         .start_with_context()
         .await;
@@ -23,23 +29,8 @@ async fn initialize_config_with_mint_and_token() {
     // Given a config account and a validator's vote account.
 
     let config = create_config(&mut context).await;
-    let vote = Keypair::new().pubkey();
-    let validator = Keypair::new().pubkey();
-
-    let mut data = vec![0u8; VoteState::size_of()];
-    // 0-4 is the version offset
-    data[4..36].copy_from_slice(validator.as_ref()); // node_pubkey
-    data[36..68].copy_from_slice(validator.as_ref()); // authorized_withdrawer
-
-    let vote_account = Account {
-        lamports: 5_000_000_000, // 5 SOL
-        data,
-        owner: solana_sdk::vote::program::ID,
-        executable: false,
-        rent_epoch: Epoch::default(),
-    };
-
-    context.set_account(&vote, &vote_account.into());
+    let validator = Pubkey::new_unique();
+    let vote = create_vote_account(&mut context, &validator, &validator);
 
     // When we initialize the stake account.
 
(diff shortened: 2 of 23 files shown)


program/src/processor/initialize_stake.rs
```diff
@@ -14,8 +23,97 @@ use crate::instruction::accounts::{Context, InitializeStakeAccounts};
...
+
+    require!(
+        config.is_initialized(),
+        ProgramError::UninitializedAccount,
+        "config"
+    );
+
+    // 2. validator_vote
+    // - owner must be the vote program
+    // - must be initialized
```
buffalojoec: The vote account is third in this instruction. Looks like it may not matter with the `Context` object, but the comments should be accurate.

Stake account is second, so you just have these two flipped.

program/src/state/stake.rs
```diff
@@ -48,4 +47,14 @@ pub struct Stake {
...
+
+    #[inline(always)]
+    pub fn is_initialized(&self) -> bool {
+        self.discriminator.as_slice() == Stake::SPL_DISCRIMINATOR_SLICE
+    }
+
+    #[inline(always)]
+    pub fn is_uninitialized(&self) -> bool {
+        self.discriminator.as_slice() == ArrayDiscriminator::UNINITIALIZED.as_slice()
+    }
```
buffalojoec: I'd recommend removing this one. You can just use `!is_initialized()` and the two are quite similar in name, which could be confusing.

program/src/state/mod.rs
```diff
@@ -9,3 +9,15 @@ pub use stake::*;
 pub fn find_vault_pda(config: &Pubkey, program_id: &Pubkey) -> (Pubkey, u8) {
     Pubkey::find_program_address(&["token-owner".as_bytes(), config.as_ref()], program_id)
 }
+
+#[inline(always)]
+pub fn find_stake_pda(validator: &Pubkey, config: &Pubkey, program_id: &Pubkey) -> (Pubkey, u8) {
+    Pubkey::find_program_address(
+        &[
+            "stake::state::stake".as_bytes(),
```
buffalojoec, line 17: Do you want this to be this whole string as bytes, or should it be `Stake::SPL_DISCRIMINATOR_SLICE`? The former is going to be longer than 8 bytes.

program/src/state/stake.rs
buffalojoec: Also keep in mind an account that hasn't been allocated yet won't even have 8 bytes of zeroes, so this could be a bit brittle if used elsewhere.

program/src/state/stake.rs
```diff
@@ -13,17 +15,14 @@ pub struct Stake {
     /// The discriminator is equal to `ArrayDiscriminator:: UNINITIALIZED` when
     /// the account is empty, and equal to `Stake::DISCRIMINATOR` when the account
     /// is initialized.
-    discriminator: [u8; 8],
+    pub discriminator: [u8; 8],
```
buffalojoec: This can probably stay private and you could roll something like a `Stake::new(..)`. What do you think?

program/src/state/stake.rs
```diff
@@ -13,17 +15,14 @@ pub struct Stake {
...
     /// Amount of staked tokens currently active
     pub amount: u64,
 
     /// Timestamp for when deactivation began. Used to judge if a given stake
     /// is inactive.
-    /// NOTE: this is a special type where all zeros means `None` to avoid
-    /// wasting space, just like `Option<NonZeroU64>`
-    // TODO: Nullable trait + PodOption?
-    pub deactivation_timestamp: UnixTimestamp,
+    pub deactivation_timestamp: Option<NonZeroU64>,
```
buffalojoec, line 28: Nice!

program/src/state/mod.rs
```diff
@@ -21,3 +21,17 @@ pub fn find_stake_pda(validator: &Pubkey, config: &Pubkey, program_id: &Pubkey)
...
 }
+
+#[inline(always)]
+pub fn get_stake_pda_signer_seeds<'a>(
+    validator: &'a Pubkey,
+    config: &'a Pubkey,
+    bump_seed: &'a [u8],
+) -> [&'a [u8]; 4] {
+    [
+        "stake::state::stake".as_bytes(),
```
buffalojoec, line 32: Same comment here about this seed. Do you want this to be this whole string as bytes, or should it be `Stake::SPL_DISCRIMINATOR_SLICE`? The former is going to be longer than 8 bytes.

clients/rust/src/pdas.rs
```diff
@@ -3,3 +3,14 @@ use solana_program::pubkey::Pubkey;
...
+pub fn find_stake_pda(validator: &Pubkey, config: &Pubkey) -> (Pubkey, u8) {
+    Pubkey::find_program_address(
+        &[
+            "stake::state::stake".as_bytes(),
+            validator.as_ref(),
+            config.as_ref(),
+        ],
+        &crate::ID,
+    )
+}
```
buffalojoec, line 16: Just curious, are you manually rolling these in the client as well as the program?

clients/rust/tests/initialize_stake.rs
```diff
@@ -80,3 +71,287 @@ async fn initialize_config_with_mint_and_token() {
...
+        &context.payer.pubkey(),
+        &stake_pda,
+        context
+            .banks_client
+            .get_rent()
+            .await
+            .unwrap()
+            .minimum_balance(Stake::LEN),
+    );
+
```
buffalojoec, line 101: Try initializing this with some arbitrary data and see if we get the same control flow. This will guard against unsolicited invocations from accounts that don't have the `Stake` discriminator.

```rust
context.set_account(
    &stake_pda,
    AccountSharedData::from(Account
        lamports: 100_000_000,
        data: vec![5; std::mem::size_of::<Stake>()],
        owner: system_program::id(),
        ..Default::default()
    }),
);
```

clients/rust/tests/initialize_stake.rs
```diff
@@ -80,3 +71,287 @@ async fn initialize_config_with_mint_and_token() {
...
+        .await
+        .unwrap_err();
+
+    // Then we expect an error.
+
+    assert_instruction_error!(err, InstructionError::AccountAlreadyInitialized);
+}
+
+#[tokio::test]
+async fn fail_initialize_stake_without_funding_account() {
```
buffalojoec: micro-nit: This test name confused me for second! How about `fail_initialize_stake_not_rent_exempt`? The current name alludes to a missing payer.

clients/rust/tests/initialize_stake.rs
```diff
@@ -80,3 +71,287 @@ async fn initialize_config_with_mint_and_token() {
...
+        .await
+        .unwrap_err();
+
+    // Then we expect an error.
+
+    assert_instruction_error!(err, InstructionError::InvalidSeeds);
+}
+
+#[tokio::test]
+async fn fail_initialize_stake_with_invalid_derivation() {
```
buffalojoec, line 145: How is this test different than the one above it?

clients/rust/tests/initialize_stake.rs
```diff
@@ -80,3 +71,287 @@ async fn initialize_config_with_mint_and_token() {
...
+
+    let validator = Pubkey::new_unique();
+    let vote = create_vote_account_with_program_id(
+        &mut context,
+        &validator,
+        &validator,
+        &system_program::ID,
+    );
+
+    // When we try initialize the stake account with an invalid derivation.
```
buffalojoec: This test isn't immediately clear what it's doing. Maybe it's the comment at the end of the block I highlighted?

Just so I understand, this is creating a stake account ahead of time, but with all-zeroes (uninitialized) and then ensuring the program fails on `UninitializedAccount`? I think technically, since the program is going to allocate, this should be `AccountAlreadyInitialized`, right? Maybe there's something I'm missing in this test.

clients/rust/tests/setup/vote.rs
```diff
@@ -0,0 +1,37 @@
...
+    context: &mut ProgramTestContext,
+    node: &Pubkey,
+    authority: &Pubkey,
+    program_id: &Pubkey,
+) -> Pubkey {
+    let vote = Pubkey::new_unique();
+    let mut data = vec![0u8; VoteState::size_of()];
+    // 0-4 is the version offset
+    data[4..36].copy_from_slice(node.as_ref()); // node_pubkey
+    data[36..68].copy_from_slice(authority.as_ref()); // authorized_withdrawer
```
buffalojoec: I'd be careful here. I know this program only cares about the first 68 bytes, but testing with accounts that don't match the real-live vote state could have unintended side effects.

Could you use a combination of `..Default::default()` and serialize?

https://github.com/anza-xyz/agave/blob/b5f5c3cdd3f9a5859c49ebc27221dc27e143d760/sdk/program/src/vote/state/mod.rs#L383-L391

buffalojoec (commented):
Looking good! I left a handful of suggestions and questions.

program/src/processor/initialize_stake.rs
febo: The numbering is not associated with the account order, it represents the steps on the validation. The reason the `validator_vote` comes first is because the we need the validator pubkey (inside the vote account) to validate the derivation of the stake account.

Perhaps we can remove the numbering if that is confusing with the account order.

program/src/state/stake.rs
febo: The issue with `!is_initialized()` is that it could be "weak", since you can pass another account of the same program and it will treat it as uninitialized because the discriminator don't match instead of checking that the discriminator is zeroed.

Maybe the naming is not idea, but what `is_initialized` and `is_uninitialied` are doing is checking the discriminator on the account.

program/src/state/mod.rs
febo: I think a `string` would more visible to clients.

program/src/state/stake.rs
febo: This type is not being used on the client, so it is private in that sense.

program/src/state/stake.rs
febo: Yeah, it worked nicely!

clients/rust/src/pdas.rs
febo: Yes, but just realised that we don't have to. We can add this to the Kinobi configuration. 🤦‍♂️

program/src/state/stake.rs
buffalojoec: I don't even see `is_uninitialized` being used in this branch. I think it makes more sense for `InitializeConfig` because the program requires the account to come pre-allocated, while this one does not.

program/src/state/mod.rs
buffalojoec: Just have to make sure it's well-documented that the account's 8-byte discriminator within its state is not the same as the seed prefix.

clients/rust/tests/initialize_stake.rs
febo: <del>This is failing on `InvalidSeeds`, since the derivation is wrong. The vote account is created with `validator` pubkey but the derivation of the `Stake` PDA is done with `Pubkey::new_unique()`. It will fail when the program does the derivation.</del>

Misread your comment. This is testing an invalid vote account, so the scenario of trying to initialize an `Stake` for an inexistent/invalid `VoteState` account. It fails with an `InvalidAccountOwner` since the account won't be owner by the correct program.

But as you pointed out, the comment is misleading, I will update it.

program/src/state/stake.rs
buffalojoec: I'll leave it up to you, but generally for a type that should always have a specific implemented constant value for a certain field, its better to guard this behind an initializer method.

clients/rust/tests/initialize_stake.rs
febo: It is not, you are right. I will remove the one above.

clients/rust/tests/setup/vote.rs
febo: Good point, I will update this.

program/src/state/stake.rs
febo: Oh, yeah, good point. It won't be used as the `Config` one. I will remove it.

program/src/state/mod.rs
febo: Good point, I will double check the comments. I think the discriminator is sort of "hidden" from clients – the client code does not need to interact with it directly.

program/src/state/stake.rs
febo: Got it, I will add a `new(..)` method on it and make the field private.

febo: What do you think about having a "manual" implementation of `Default` instead of the `new(..)` method:
```rust
impl Default for Stake {
    fn default() -> Self {
        Self {
            discriminator: Stake::SPL_DISCRIMINATOR.into(),
            amount: u64::default(),
            deactivation_timestamp: Option::default(),
            deactivating_amount: u64::default(),
            inactive_amount: u64::default(),
            authority: Pubkey::default(),
            validator: Pubkey::default(),
            last_seen_holder_rewards: u64::default(),
            last_seen_stake_rewards: u64::default(),
        }
    }
}
```

febo: Hmmm, never mind, it does not work.

buffalojoec: Ahh nuts. I think `new` is okay, where any of the fields you plan to initialize with, say `0`, are just that, and it takes the minimum necessary initialization values.

clients/rust/src/pdas.rs
buffalojoec: Okay, that can be done in a separate change, imo.

program/src/state/mod.rs
febo: Same as comment above, will make sure it is documented but clients should not need to interact with the discriminator.

clients/rust/tests/initialize_stake.rs
febo: Did you see this comment?
```
let account = context.banks_client.get_account(stake_pda).await.unwrap();
// TODO: Shouldn't we get an error from the instruction?
assert!(account.is_none());
```
The transaction succeeds but the account is not created. Is this because of how the tests are run using `banks_client`?

buffalojoec: Ahh I also noticed this in my tests! I have a note to check into it, but yeah I'm assuming it has to do with the way CPI is stubbed in `program-test`. Maybe drop this test for now?

clients/rust/tests/initialize_stake.rs
buffalojoec: The error thrown in this test is `UninitializedAccount`.

clients/rust/tests/initialize_stake.rs
febo: This will guard from pre-allocated (or initialized) accounts, right? The program asserts that `stake.data_is_empty()` before trying to initialize the account – this would be covered by the `fail_initialize_stake_with_initialized_account` test.

I think this "manual" initialization can be added to other instructions where we need to assert that the account is initialized, so the discriminator must match.

clients/rust/tests/setup/vote.rs
febo: Looks much better now, thanks!

buffalojoec (approved):
Awesome! Ship it! 🚀

clients/rust/tests/initialize_stake.rs
buffalojoec: > The program asserts that `stake.data_is_empty()` before trying to initialize the account – this would be covered by the `fail_initialize_stake_with_initialized_account` test.

`fail_initialize_stake_with_initialized_account` is the test I left this comment on. Sorry, I think I maybe grabbed the wrong lines here - it should have been lower down.

I'm basically suggesting that instead of using the `InitializeStake` instruction to initialize the account, then using `InitializeStake` again to verify it won't re-initialize, you could set up the account with arbitrary data, to ensure that data initialized without the `Stake` discriminator (or even all zeroes) still fails on `stake.data_is_empty()`.

As I type it out, maybe it's not all that useful, but I'll leave it up to you.

febo: I will add it to another instruction, one that does not have the `stake.data_is_empty()` so we assert against the discriminator.

program/src/processor/initialize_stake.rs
```diff
@@ -10,12 +18,101 @@ use crate::instruction::accounts::{Context, InitializeStakeAccounts};
...
+
+    let data = &ctx.accounts.validator_vote.try_borrow_data()?;
+
+    require!(
+        VoteState::is_correct_size_and_initialized(data),
+        ProgramError::InvalidAccountData,
+        "validator_vote"
+    );
+
+    let validator = Pubkey::from(*array_ref!(data, 4, 32));
```
joncinque, line 69: Sorry that this wasn't clear -- let's use the validator vote address instead of the node id, since the node id can be reused in multiple vote accounts, which could easily let someone impersonate another node

joncinque (commented):
Just a fly-by comment, the rest looks good!

program/src/processor/initialize_stake.rs
febo: Created a PR for it: #30

PR #21, opened 2024-07-03
Processor: `DeactivateStake`

febo:
This PR adds the processor for the `DeactivateStake` instruction.

### DIFF
--- a/clients/rust/tests/deactivate_stake.rs
+++ b/clients/rust/tests/deactivate_stake.rs
@@ -1,0 +1,375 @@
+#![cfg(feature = "test-sbf")]
+
+mod setup;
+
+use borsh::BorshSerialize;
+use paladin_stake::{
+    accounts::Stake, errors::StakeError, instructions::DeactivateStakeBuilder, pdas::find_stake_pda,
+};
+use setup::{config::create_config, stake::create_stake, vote::create_vote_account};
+use solana_program_test::{tokio, ProgramTest};
+use solana_sdk::{
+    account::{Account, AccountSharedData},
+    instruction::InstructionError,
+    pubkey::Pubkey,
+    signature::{Keypair, Signer},
+    transaction::Transaction,
+};
+
+#[tokio::test]
+async fn deactivate_stake() {
+    let mut context = ProgramTest::new("stake_program", paladin_stake::ID, None)
+        .start_with_context()
+        .await;
+
+    // Given a config account and a validator's vote account.
+
+    let config = create_config(&mut context).await;
+    let validator = Pubkey::new_unique();
+    let authority = Keypair::new();
+    let vote = create_vote_account(&mut context, &validator, &authority.pubkey()).await;
+
+    // And a stake account.
+
+    let stake_pda = create_stake(&mut context, &validator, &vote, &config).await;
+
+    let account = get_account!(context, stake_pda);
+    let mut stake_account = Stake::from_bytes(account.data.as_ref()).unwrap();
+    // "manually" set the amount to 100
+    stake_account.amount = 100;
+
+    let updated_stake = Account {
+        lamports: account.lamports,
+        data: stake_account.try_to_vec().unwrap(),
+        owner: account.owner,
+        executable: account.executable,
+        rent_epoch: account.rent_epoch,
+    };
+
+    context.set_account(&stake_pda, &updated_stake.into());
+
+    // When we deactivate an amount from the stake account.
+
+    let deactivate_ix = DeactivateStakeBuilder::new()
+        .stake(stake_pda)
+        .stake_authority(authority.pubkey())
+        .args(50)
+        .instruction();
+
+    let tx = Transaction::new_signed_with_payer(
+        &[deactivate_ix],
+        Some(&context.payer.pubkey()),
+        &[&context.payer, &authority],
+        context.last_blockhash,
+    );
+    context.banks_client.process_transaction(tx).await.unwrap();
+
+    // Then the deactivation should be successful.
+
+    let account = get_account!(context, stake_pda);
+    let stake_account = Stake::from_bytes(account.data.as_ref()).unwrap();
+
+    assert_eq!(stake_account.amount, 100);
+    assert_eq!(stake_account.deactivating_amount, 50);
+    assert!(stake_account.deactivation_timestamp.value().is_some())
+}
+
+#[tokio::test]
+async fn deactivate_stake_with_active_deactivation() {
+    let mut context = ProgramTest::new("stake_program", paladin_stake::ID, None)
+        .start_with_context()
+        .await;
+
+    // Given a config account and a validator's vote account.
+
+    let config = create_config(&mut context).await;
+    let validator = Pubkey::new_unique();
+    let authority = Keypair::new();
+    let vote = create_vote_account(&mut context, &validator, &authority.pubkey()).await;
+
+    // And a stake account.
+
+    let stake_pda = create_stake(&mut context, &validator, &vote, &config).await;
+
+    let account = get_account!(context, stake_pda);
+    let mut stake_account = Stake::from_bytes(account.data.as_ref()).unwrap();
+    // "manually" set the amount to 100
+    stake_account.amount = 100;
+
+    let updated_stake = Account {
+        lamports: account.lamports,
+        data: stake_account.try_to_vec().unwrap(),
+        owner: account.owner,
+        executable: account.executable,
+        rent_epoch: account.rent_epoch,
+    };
+
+    context.set_account(&stake_pda, &updated_stake.into());
+
+    // And we deactivate an amount from the stake account.
+
+    let deactivate_ix = DeactivateStakeBuilder::new()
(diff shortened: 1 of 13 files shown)


febo:
@joncinque @buffalojoec The implementation of the instruction is complete, but I moved it to draft since I remembered that we have the `max_deactivation_basis_points` config. Is this the limit that can be deactivated at once?

If yes, then we need to pass the `Config` account to this instruction – currently we don't have a way to check that.

buffalojoec:
> @joncinque @buffalojoec The implementation of the instruction is complete, but I moved it to draft since I remembered that we have the `max_deactivation_basis_points` config. Is this the limit that can be deactivated at once?

Looks like it, yep. It says it's a proportion... I'm assuming of the `token_amount_delegated` field?

> If yes, then we need to pass the `Config` account to this instruction – currently we don't have a way to check that.

Yeah, makes sense to me!

febo:
> > @joncinque @buffalojoec The implementation of the instruction is complete, but I moved it to draft since I remembered that we have the `max_deactivation_basis_points` config. Is this the limit that can be deactivated at once?
> 
> Looks like it, yep. It says it's a proportion... I'm assuming of the `token_amount_delegated` field?

Yeah, I am using the `token_amount_delegated` field. The other option is to used the staked amount of the validator, but that would create edge cases where multiple deactivations would be needed to deactivate the whole stake of a validator, with increasingly lower values.

But one thing to keep in mind is that the `token_amount_delegated` is not updated on deactivation, only when the stake is moved to inactive. When multiple requests for deactivation happen within the "cooldown time", they will be evaluated using the same `token_amount_delegated` value.

program/src/processor/deactivate_stake.rs
```diff
@@ -1,21 +1,136 @@
...
 /// Instruction data: amount of tokens to deactivate, as a little-endian u64
 pub fn process_deactivate_stake(
-    _program_id: &Pubkey,
-    _ctx: Context<DeactivateStakeAccounts>,
-    _amount: u64,
+    program_id: &Pubkey,
+    ctx: Context<DeactivateStakeAccounts>,
+    amount: u64,
 ) -> ProgramResult {
+    // Account valuidation.
```
buffalojoec: Sorry! Spotted this on both PR's. 😁

```suggestion
    // Account validation.
```

program/src/processor/deactivate_stake.rs
```diff
@@ -1,21 +1,136 @@
...
+};
+
+/// Helper to calculate the maximum amount that can be deactivated.
+#[inline(always)]
+fn get_max_deactivation_amount(total_amount: u64, basis_points: u16) -> Result<u64, ProgramError> {
+    total_amount
+        .checked_mul(basis_points as u64)
+        .ok_or(ProgramError::ArithmeticOverflow)?
+        .checked_div(10_000)
+        .ok_or(ProgramError::ArithmeticOverflow)
```
buffalojoec: nit: Just a tad cleaner, if you want!

```suggestion
    total_amount
        .checked_mul(basis_points as u64)
        .and_then(|p| p.checked_div(10_000))
        .ok_or(ProgramError::ArithmeticOverflow)
```

clients/rust/tests/deactivate_stake.rs
```diff
@@ -0,0 +1,493 @@
...
+        context.last_blockhash,
+    );
+    context.banks_client.process_transaction(tx).await.unwrap();
+
+    // Then the deactivation should be "updated".
+
+    let account = get_account!(context, stake_pda);
+    let stake_account = Stake::from_bytes(account.data.as_ref()).unwrap();
+    assert_eq!(stake_account.deactivating_amount, 1);
+    assert!((timestamp as u64) < stake_account.deactivation_timestamp.value().unwrap());
```
buffalojoec: Why not just create a variable with the advanced timestamp value (ie. `timestamp + 1000`), and do a direct comparison here?

clients/rust/tests/deactivate_stake.rs
```diff
@@ -0,0 +1,493 @@
...
+    );
+    let err = context
+        .banks_client
+        .process_transaction(tx)
+        .await
+        .unwrap_err();
+
+    // Then we expect an error.
+
+    assert_instruction_error!(err, InstructionError::UninitializedAccount);
```
buffalojoec, line 432: Maybe also add a case for an uninitialized config account, too?

clients/rust/tests/deactivate_stake.rs
```diff
@@ -0,0 +1,493 @@
...
+
+#[tokio::test]
+async fn fail_deactivate_stake_with_uninitialized_stake_account() {
+    let mut context = ProgramTest::new("stake_program", paladin_stake::ID, None)
+        .start_with_context()
+        .await;
+
+    // Given a config account and a validator's vote account.
+
+    let config = create_config(&mut context).await;
```
buffalojoec: It's not clear in this test exactly _how_ you're breaching the maximum deactivation proportion. The value is hard-coded to 500 basis points over in the setup. Might be nice to (re)configure this value manually over here, or at least add a comment for it?

buffalojoec (commented):
Nice, this one is really close. Just a few small things to tune up.

program/src/processor/deactivate_stake.rs
febo: Nice!

clients/rust/tests/deactivate_stake.rs
febo: Sorry, I don't follow your comment. What would this variable replace?

clients/rust/tests/deactivate_stake.rs
febo: Done.

clients/rust/tests/deactivate_stake.rs
febo: Good point, added a helper that takes the values as arguments and added a comment.

clients/rust/tests/deactivate_stake.rs
buffalojoec: You're checking that the timestamp you first set on the stake account is less than the timestamp that's currently on the stake account. The check is okay, I'm just suggesting being a bit more explicit about checking the stake account's value.

```rust
let original_timestamp = clock.timestamp;
let new_timestamp = original_timestamp.saturating_add(1000);
clock.timestamp = new_timestamp;
/* ... */
assert_eq!(stake_account.timestamp, new_timestamp);
```

febo: Got it, makes sense. I updated the test.

buffalojoec (approved):
ship it

program/src/processor/deactivate_stake.rs
```diff
@@ -1,21 +1,135 @@
...
+};
+
+/// Helper to calculate the maximum amount that can be deactivated.
+#[inline(always)]
+fn get_max_deactivation_amount(total_amount: u64, basis_points: u16) -> Result<u64, ProgramError> {
+    total_amount
+        .checked_mul(basis_points as u64)
+        .and_then(|p| p.checked_div(10_000))
+        .ok_or(ProgramError::ArithmeticOverflow)
+}
```
joncinque, line 22: for highly staked validators and a large `basis_points`, this could potentially overflow -- can you widen this to a `u128`? Using SOL numbers, if `basis_points` is set to 10,000 (can deactivate the whole amount all at once), and 500 million SOL (500 * 10 ^ 6 * 10 ^ 9), you'd quickly blow past `u64::MAX`

program/src/processor/deactivate_stake.rs
```diff
@@ -1,21 +1,135 @@
...
+
+    require!(
+        ctx.accounts.stake_authority.key == &stake.authority,
+        StakeError::InvalidAuthority,
+        "stake_authority"
+    );
+
+    // Validate the amount.
+
+    require!(amount > 0, StakeError::InvalidAmount);
```
joncinque, line 110: I'm wondering if it's worth allowing 0 -- that means someone can cancel a deactivation. What do you think?

joncinque (commented):
Looks good for the most part! Just a couple of flyby suggestions

program/src/processor/deactivate_stake.rs
febo: Hmmm, that would look like a side-effect so not sure whether we should allow that or create an specific instruction for that. Also, there is no "effect" of deactivating unless you move the tokens to inactive.

joncinque: Moving the tokens to inactive is a permissionless instruction, so if I accidentally deactivate, then I need to be ready to re-activate right when the time passes if I don't want to lose out on any rewards. With the current behavior, I would probably override my incorrect deactivation with another deactivation for 1 token, which is kind of silly.

If you want to add a new instruction, that's fine by me! But that would be a many-line change vs a one-line change.

febo: Good point, I will change so an amount of `0` cancels the deactivation.

PR #22, opened 2024-07-03
Processor: `InactivateStake`

febo:
This PR adds the processor for the`InactivateStake` instruction.

### DIFF
--- a/clients/rust/tests/inactivate_stake.rs
+++ b/clients/rust/tests/inactivate_stake.rs
@@ -93,12 +93,12 @@ async fn inactivate_stake() {
     assert_eq!(stake_account.inactive_amount, 50);
     assert!(stake_account.deactivation_timestamp.value().is_none());
 
-    // And the total delegated on the config must decrease.
+    // And the total delegated on the config stays the same.
 
     let account = get_account!(context, config);
     let config_account = Config::from_bytes(account.data.as_ref()).unwrap();
 
-    assert_eq!(config_account.token_amount_delegated, 50);
+    assert_eq!(config_account.token_amount_delegated, 100);
 }
 
 #[tokio::test]
--- a/clients/rust/tests/inactivate_stake.rs
+++ b/clients/rust/tests/inactivate_stake.rs
@@ -102,7 +102,7 @@ async fn inactivate_stake() {
 }
 
 #[tokio::test]
-async fn inactivate_stake_with_no_deactivated_amount() {
+async fn fail_inactivate_stake_with_no_deactivated_amount() {
     let mut context = ProgramTest::new("stake_program", paladin_stake::ID, None)
         .start_with_context()
         .await;
@@ -147,25 +147,15 @@ async fn inactivate_stake_with_no_deactivated_amount() {
         &[&context.payer],
         context.last_blockhash,
     );
-    context.banks_client.process_transaction(tx).await.unwrap();
+    let err = context
+        .banks_client
+        .process_transaction(tx)
+        .await
+        .unwrap_err();
 
-    // Then the tranasction succeeds and the stake account should
-    // remain the same.
+    // Then we expect an error.
 
-    let account = get_account!(context, stake_pda);
-    let stake_account = Stake::from_bytes(account.data.as_ref()).unwrap();
-
-    assert_eq!(stake_account.amount, 100);
-    assert_eq!(stake_account.deactivating_amount, 0);
-    assert_eq!(stake_account.inactive_amount, 0);
-    assert!(stake_account.deactivation_timestamp.value().is_none());
-
-    // And the total delegated on the config is the same.
-
-    let account = get_account!(context, config);
-    let config_account = Config::from_bytes(account.data.as_ref()).unwrap();
-
-    assert_eq!(config_account.token_amount_delegated, 100);
+    assert_custom_error!(err, StakeError::NoDeactivatedTokens);
 }
 
 #[tokio::test]
--- a/clients/rust/tests/inactivate_stake.rs
+++ b/clients/rust/tests/inactivate_stake.rs
@@ -5,11 +5,16 @@ mod setup;
 use borsh::BorshSerialize;
 use paladin_stake::{
     accounts::{Config, Stake},
+    errors::StakeError,
     instructions::InactivateStakeBuilder,
     pdas::find_stake_pda,
     NullableU64,
 };
-use setup::{config::create_config, stake::create_stake, vote::create_vote_account};
+use setup::{
+    config::{create_config, create_config_with_args},
+    stake::create_stake,
+    vote::create_vote_account,
+};
 use solana_program_test::{tokio, ProgramTest};
 use solana_sdk::{
     account::{Account, AccountSharedData},
@@ -30,18 +35,12 @@ async fn inactivate_stake() {
 
     let config = create_config(&mut context).await;
     // "manually" set the total amount delegated
-    let account = get_account!(context, config);
+    let mut account = get_account!(context, config);
     let mut config_account = Config::from_bytes(account.data.as_ref()).unwrap();
     config_account.token_amount_delegated = 100;
-
-    let updated_config = Account {
-        lamports: account.lamports,
-        data: config_account.try_to_vec().unwrap(),
-        owner: account.owner,
-        executable: account.executable,
-        rent_epoch: account.rent_epoch,
-    };
-    context.set_account(&config, &updated_config.into());
+    // "manually" update the config account data
+    account.data = config_account.try_to_vec().unwrap();
+    context.set_account(&config, &account.into());
 
     // And a stake account (amount = 100).
 
@@ -51,7 +50,7 @@ async fn inactivate_stake() {
 
     let stake_pda = create_stake(&mut context, &validator, &vote, &config).await;
 
-    let account = get_account!(context, stake_pda);
+    let mut account = get_account!(context, stake_pda);
     let mut stake_account = Stake::from_bytes(account.data.as_ref()).unwrap();
     // "manually" set the stake values
     stake_account.amount = 100;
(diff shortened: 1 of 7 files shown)


program/src/processor/inactivate_stake.rs
```diff
@@ -13,8 +21,91 @@ use crate::instruction::accounts::{Context, InactivateStakeAccounts};
 /// 0. `[w]` Stake config account
 /// 1. `[w]` Validator stake account
 pub fn process_inactivate_stake(
-    _program_id: &Pubkey,
-    _ctx: Context<InactivateStakeAccounts>,
+    program_id: &Pubkey,
+    ctx: Context<InactivateStakeAccounts>,
 ) -> ProgramResult {
+    // Account valuidation.
```
buffalojoec:
```suggestion
    // Account validation.
```

program/src/processor/inactivate_stake.rs
```diff
@@ -13,8 +21,91 @@ use crate::instruction::accounts::{Context, InactivateStakeAccounts};
...
+    if let Some(timestamp) = stake.deactivation_timestamp {
+        let current = Clock::get()?
+            .unix_timestamp
+            .saturating_sub(config.cooldown_time_seconds) as u64;
+
+        // TODO: should we fail or just do nothing?
+        require!(
+            current >= timestamp.get(),
+            StakeError::ActiveDeactivationCooldown
+        );
```
buffalojoec: In my opinion this should be an error. A successful run of this program instruction is successful inactivation of stake. This would be an error case to me.

buffalojoec: Before throwing, I wonder if we should emit some kind of log to tell users how much time is remaining in cooldown. What do you think?

program/src/processor/inactivate_stake.rs
```diff
@@ -13,8 +21,91 @@ use crate::instruction::accounts::{Context, InactivateStakeAccounts};
...
+        ctx.accounts.stake.key == &derivation,
+        ProgramError::InvalidSeeds,
+        "stake",
+    );
+
+    // Inactivates the stake if elegible.
+    //
+    // Note: we do not fail if there are no tokens to inactivate
+
+    if let Some(timestamp) = stake.deactivation_timestamp {
```
buffalojoec: I also think this should be an error, for similar reasons as my other comment. A user who tries to inactivate tokens when there are none to inactivate should see an error that tells them exactly that, in my opinion. To me, successfully returning the program without logs or errors seems misleading here.

program/src/processor/inactivate_stake.rs
```diff
@@ -13,8 +21,91 @@ use crate::instruction::accounts::{Context, InactivateStakeAccounts};
...
+        require!(
+            current >= timestamp.get(),
+            StakeError::ActiveDeactivationCooldown
+        );
+
+        msg!("Deactivating {} token(s)", stake.deactivating_amount);
+
+        config.token_amount_delegated = config
+            .token_amount_delegated
+            .saturating_sub(stake.deactivating_amount);
```
buffalojoec: Just wondering, is this supposed to be updated here? It says this field is the "total number of tokens delegated to the system". Does this mean `amount + inactive_amount`? Should this field instead be updated on `WithdrawInactiveStake`?

clients/rust/tests/inactivate_stake.rs
```diff
@@ -0,0 +1,106 @@
...
+    stake_account.deactivation_timestamp = NullableU64::from(timestamp as u64);
+
+    let updated_stake = Account {
+        lamports: account.lamports,
+        data: stake_account.try_to_vec().unwrap(),
+        owner: account.owner,
+        executable: account.executable,
+        rent_epoch: account.rent_epoch,
+    };
+    context.set_account(&stake_pda, &updated_stake.into());
```
buffalojoec: nit: Can you just do `let mut account = get_account!(/* ... */);` then modify the account itself in-place? Might make these tests a little clearer.

clients/rust/tests/inactivate_stake.rs
```diff
@@ -96,11 +98,213 @@ async fn inactivate_stake() {
...
+    let tx = Transaction::new_signed_with_payer(
+        &[inactivate_ix],
+        Some(&context.payer.pubkey()),
+        &[&context.payer],
+        context.last_blockhash,
+    );
+    context.banks_client.process_transaction(tx).await.unwrap();
+
+    // Then the tranasction succeeds and the stake account should
+    // remain the same.
```
buffalojoec: As per my other comments on the processor, I definitely feel it makes sense to have an error here.

clients/rust/tests/inactivate_stake.rs
```diff
@@ -96,11 +98,213 @@ async fn inactivate_stake() {
...
+    let validator = Pubkey::new_unique();
+    let authority = Keypair::new();
+    let vote = create_vote_account(&mut context, &validator, &authority.pubkey()).await;
+
+    let stake_pda = create_stake(&mut context, &validator, &vote, &config).await;
+
+    let account = get_account!(context, stake_pda);
+    let mut stake_account = Stake::from_bytes(account.data.as_ref()).unwrap();
+    // "manually" set the stake values
+    stake_account.amount = 100;
```
buffalojoec, line 189: micro-nit: you aren't setting deactivating amount here. Even though `InvalidSeeds` will throw first, just wanted to point it out.

buffalojoec (commented):
Looking really good! I left a few suggestions and weighed in on some of your open questions.

program/src/processor/inactivate_stake.rs
febo: Added a log message to show the remaining seconds for deactivation.

program/src/processor/inactivate_stake.rs
febo: One advantage of not throwing an error is that you could batch inactivations without having to filter them. This way you could schedule a service that run through all accounts, like a "clean-up" process.

febo: Hmmm, but for this to work we also should not throw an error if the cooldown time hasn't being reached.

program/src/processor/inactivate_stake.rs
febo: That is a good question. My interpretation was to decrease the total once the tokens are "moved" to "inactive" – although they only really move on the `WithdrawInactiveStake` as you pointed out. So the `inactive` tokens are not eligible for rewards independently from whether they have been withdrawn or not.

If we don't decrease the `token_amount_delegated`, then we should not decrease the `amount` either.

clients/rust/tests/inactivate_stake.rs
febo: Done.

clients/rust/tests/inactivate_stake.rs
febo: Updated.

program/src/processor/inactivate_stake.rs
febo: Had another look at this and we could do the following:
- `config.token_amount_delegated`: it is the total number of tokens (`amount` + `inactive_amount`), only decreases on `WithdrawInactiveStake`.
- `stake.amount`: it is only the "active" amount, it decreases when tokens are "moved" from deactivated to inactive.

buffalojoec: I think this is correct yeah, since `WithdrawInactiveStake` is documented as saying it's moving tokens "out of the vault", which is represented by that `token_amount_delegated`.

program/src/processor/inactivate_stake.rs
febo: It is now throwing an error.

buffalojoec (approved):
Nice work!!

program/src/error.rs
```diff
@@ -64,6 +64,14 @@ pub enum StakeError {
     /// 13 - Amount exeeds maximum deactivation amount
     #[error("Amount exeeds maximum deactivation amount")]
     MaximumDeactivationAmountExceeded,
+
+    /// 14 - Active deactivation cooldown
+    #[error("Active deactivation cooldown")]
+    ActiveDeactivationCooldown,
+
+    /// 16 - No deactivated tokens
```
joncinque, line 72: nit: this should be 15

program/src/processor/inactivate_stake.rs
```diff
@@ -13,8 +21,90 @@ use crate::instruction::accounts::{Context, InactivateStakeAccounts};
...
+        let current = Clock::get()?
+            .unix_timestamp
+            .saturating_sub(config.cooldown_time_seconds) as u64;
+
+        require!(
+            current >= timestamp.get(),
+            StakeError::ActiveDeactivationCooldown,
+            "{} second(s) remaining for deactivation",
+            timestamp.get().saturating_sub(current),
+        );
```
joncinque, line 92: nit: this logic is a little confusing and looks backwards, since `current` isn't the current timestamp. It's up to you, but this would be clearer to me if you do something like:

```
let inactive_timestamp = timestamp.get().saturating_add(config.cooldown_time_seconds);
require!(Clock::get()?.unix_timestamp >= inactive_timestamp);
```

program/src/processor/inactivate_stake.rs
```diff
@@ -13,8 +21,90 @@ use crate::instruction::accounts::{Context, InactivateStakeAccounts};
...
+
+        msg!("Deactivating {} token(s)", stake.deactivating_amount);
+
+        // moves deactivating amount to inactive
+        stake.amount = stake.amount.saturating_sub(stake.deactivating_amount);
```
joncinque, line 97: This is the paranoid part of me, but how about doing a `checked_sub` instead? This should never fail, right?

program/src/processor/inactivate_stake.rs
```diff
@@ -13,8 +21,90 @@ use crate::instruction::accounts::{Context, InactivateStakeAccounts};
...
+        stake.inactive_amount = stake
+            .inactive_amount
+            .saturating_add(stake.deactivating_amount);
```
joncinque, line 100: Same here, we should be able to safely do checked math, right?

joncinque (commented):
Just a few more flyby comments, these are looking good!

program/src/processor/inactivate_stake.rs
febo: I will change it to `checked_sub` – and yes, this should never fail since the `deactivating_amount` will always be smaller or equal to the amount.

program/src/processor/inactivate_stake.rs
febo: Yep!

program/src/processor/inactivate_stake.rs
febo: Agree, your suggestion makes the logic more clear.

PR #23, opened 2024-07-09
Processor: `StakeTokens`

febo:
This PR adds the processor for the `StakeTokens` instruction.


clients/rust/tests/fixtures/paladin_rewards_program.so
buffalojoec, line 1: 😅

program/src/processor/stake_tokens.rs
```diff
@@ -12,21 +25,167 @@ use crate::instruction::accounts::{Context, StakeTokensAccounts};
...
-/// 4. `[]` Validator vote account
-/// 3. `[]` Stake Token Mint
-/// 4. `[]` Stake Token Vault, to hold all staked tokens.
-///   Must be the token account on the stake config account
-/// 5. `[]` Token program
+/// 4. `[ ]` Validator vote account
+/// 3. `[ ]` Stake Token Mint
+/// 4. `[ ]` Stake Token Vault, to hold all staked tokens
+///          (must be the token account on the stake config account)
+/// 5. `[ ]` Token program
```
buffalojoec: The vault needs to be writable here, and your numbers are a bit jumbled up!

clients/rust/tests/setup/stake.rs
```diff
@@ -2,7 +2,70 @@ use paladin_stake_program_client::{
...
+}
+
+impl StakeManager {
+    pub async fn new(context: &mut ProgramTestContext, config: &Pubkey) -> Self {
+        let mut manager = Self {
+            stake: Pubkey::default(),
+            authority: Keypair::new(),
+            validator: Pubkey::new_unique(),
+            vote: Pubkey::default(),
+        };
```
buffalojoec: nit: why not just initialize variables as you create the keys, rather than generating these ones that you're going to immediately discard?

```rust
let vote = create_vote_account(/* ... */);
let stake = create_vote_account(/* ... */);

Self {
    stake,
    vote,
}
```

program/src/processor/stake_tokens.rs
```diff
@@ -12,21 +25,167 @@ use crate::instruction::accounts::{Context, StakeTokensAccounts};
...
+
+    let mut account_infos: Vec<AccountInfo> = Vec::with_capacity(5 + ctx.remaining_accounts.len());
+    account_infos.push(ctx.accounts.token_program.clone());
+    account_infos.push(ctx.accounts.source_token_account.clone());
+    account_infos.push(ctx.accounts.mint.clone());
+    account_infos.push(ctx.accounts.vault.clone());
+    account_infos.push(ctx.accounts.token_account_authority.clone());
+    account_infos.extend_from_slice(ctx.remaining_accounts);
+
+    invoke(&transfer_ix, &account_infos)
```
buffalojoec: You can use Token-2022's helper for this!
https://github.com/paladin-bladesmith/lockup-program/blob/52849fc8b0878fa76f37d6accad06723667d9268/program/src/processor.rs#L114-L124

program/src/processor/stake_tokens.rs
```diff
@@ -12,21 +25,167 @@ use crate::instruction::accounts::{Context, StakeTokensAccounts};
...
+        decimals,
+    )?;
+
+    ctx.remaining_accounts.iter().for_each(|account| {
+        transfer_ix.accounts.push(AccountMeta {
+            is_signer: account.is_signer,
+            is_writable: account.is_writable,
+            pubkey: *account.key,
+        });
+    });
```
buffalojoec: Same here for the helper.

buffalojoec (commented):
This is looking great, but I think the test environment is missing the transfer hook in the mint and token accounts, isn't it? You use the helper to build the instruction, but you're not initializing the extra metas for the rewards program, nor the extensions in the mint and token accounts.

Let me know if I'm just missing it, but otherwise we need to get the whole mint configuration set up so we can make sure staking goes off without a hitch with the transfer hook configs.

Side note: This is another reason I'm still salty about not using a monorepo. Now we have to reimplement all of the transfer hook configurations here in the stake program's test environment, which is subject to error. Perhaps we should instead move a few things into the rewards program's client-side (like `get_extra_metas()`) and extract some of the test setup (like `setup_mint` and `setup_holder_rewards`). What do you think?

febo:
> This is looking great, but I think the test environment is missing the transfer hook in the mint and token accounts, isn't it? You use the helper to build the instruction, but you're not initializing the extra metas for the rewards program, nor the extensions in the mint and token accounts.

I think they are been set, unless I am missing something.
* transfer hook initialization: https://github.com/paladin-bladesmith/stake-program/blob/898fd4bb752264ce379f03f43c0f981149aff51a/clients/rust/tests/setup/token.rs#L67-L77
* extra metas: https://github.com/paladin-bladesmith/stake-program/blob/898fd4bb752264ce379f03f43c0f981149aff51a/clients/rust/tests/setup/rewards.rs#L69-L73

Token accounts are ATAs, so transfer hook is already set on them by ATA program; extra metas account is initialized on `InitializeHolderRewards` from Rewards program (the account is funded on the tests).

program/src/processor/stake_tokens.rs
febo: Nice! I was thinking that there should have been a helper for this.

buffalojoec:
> I think they are been set, unless I am missing something.

Oops, yeah looks like I missed that then!

program/src/processor/stake_tokens.rs
febo: This looks way better:
```rust
spl_token_2022::onchain::invoke_transfer_checked(
   &spl_token_2022::ID,
   ctx.accounts.source_token_account.clone(),
   ctx.accounts.mint.clone(),
   ctx.accounts.vault.clone(),
   ctx.accounts.token_account_authority.clone(),
   ctx.remaining_accounts,
   amount,
   decimals,
   &[],
)
```

clients/rust/tests/setup/stake.rs
febo: Good call!

buffalojoec (approved):

program/src/processor/stake_tokens.rs
```diff
@@ -12,21 +23,150 @@ use crate::instruction::accounts::{Context, StakeTokensAccounts};
...
+///          (PDA seeds: ['stake', validator, config_account])
 /// 2. `[w]` Token Account
 /// 3. `[s]` Owner or delegate of the token account
-/// 4. `[]` Validator vote account
-/// 3. `[]` Stake Token Mint
-/// 4. `[]` Stake Token Vault, to hold all staked tokens.
-///   Must be the token account on the stake config account
-/// 5. `[]` Token program
-/// 6.. Extra accounts required for the transfer hook
+/// 4. `[ ]` Validator vote account
```
joncinque: Is this needed in the end since we can derive from the stake account data?

joncinque (approved):
Beautiful!

program/src/processor/stake_tokens.rs
febo: Good point, I removed the account.

joncinque (approved):

PR #24, opened 2024-07-11
Processor: `WithdrawInactiveStake`

febo:
This PR adds the processor for the `WithdrawInactiveStake` instruction.


clients/rust/tests/setup/rewards.rs
```diff
@@ -88,7 +88,11 @@ pub async fn create_holder_rewards_pool(
         context.last_blockhash,
     );
 
-    context.banks_client.process_transaction(tx).await.unwrap();
+    context
+        .banks_client
+        .process_transaction_with_metadata(tx)
```
buffalojoec: y?

program/src/state/mod.rs
```diff
@@ -4,11 +4,20 @@ pub mod stake;
...
 
+#[inline(always)]
+pub fn create_vault_pda<'a>(
+    config: &'a Pubkey,
+    bump_seed: &'a [u8],
+    program_id: &'a Pubkey,
+) -> Result<Pubkey, PubkeyError> {
+    Pubkey::create_program_address(&get_vault_pda_signer_seeds(config, bump_seed), program_id)
+}
+
```
buffalojoec, line 20: Just curious why I might use this instead of `find_vault_pda`? If I have to provide the bump seed here, that would mean I already found an off-curve address, so why run the hash again?

program/src/processor/withdraw_inactive_stake.rs
```diff
@@ -17,10 +30,167 @@ use crate::instruction::accounts::{Context, WithdrawInactiveStakeAccounts};
...
+
+    let mut account_infos: Vec<AccountInfo> = Vec::with_capacity(5 + ctx.remaining_accounts.len());
+    account_infos.push(ctx.accounts.token_program.clone());
+    account_infos.push(ctx.accounts.vault.clone());
+    account_infos.push(ctx.accounts.mint.clone());
+    account_infos.push(ctx.accounts.destination_token_account.clone());
+    account_infos.push(ctx.accounts.vault_authority.clone());
+    account_infos.extend_from_slice(ctx.remaining_accounts);
+
+    invoke_signed(&transfer_ix, &account_infos, &[&signer_seeds])
```
buffalojoec: Again just plugging the Token-2022 onchain helper!
https://github.com/paladin-bladesmith/lockup-program/blob/52849fc8b0878fa76f37d6accad06723667d9268/program/src/processor.rs#L114-L125

program/src/processor/withdraw_inactive_stake.rs
```diff
@@ -17,10 +30,167 @@ use crate::instruction::accounts::{Context, WithdrawInactiveStakeAccounts};
...
+        ctx.accounts.stake_authority.key == &stake.authority,
+        StakeError::InvalidAuthority,
+        "stake authority",
+    );
+
+    // vault authority
+    // - derivation must match
+
+    let bump = [config.signer_bump];
+    let vault_signer = create_vault_pda(ctx.accounts.config.key, &bump, program_id)?;
```
buffalojoec: Yeah, here also confuses me with the `create_` vs. `find_`. Why are we passing the config account's bump seed to create the vault PDA? Isn't there a chance the true vault bump seed could be different?

program/src/processor/withdraw_inactive_stake.rs
```diff
@@ -17,10 +30,167 @@ use crate::instruction::accounts::{Context, WithdrawInactiveStakeAccounts};
...
+
+    let mint_data = ctx.accounts.mint.try_borrow_data()?;
+    let mint = PodStateWithExtensions::<PodMint>::unpack(&mint_data)?;
+    let decimals = mint.base.decimals;
+
+    // Update the config and stake account data.
+
+    require!(
+        amount <= stake.inactive_amount,
+        StakeError::NoInactivatedTokens,
```
buffalojoec: nit: This could be `NotEnoughInactivatedTokens` or something like that, rather than no/none.

buffalojoec (commented):
Looks good! Just a couple small comments.

clients/rust/tests/setup/rewards.rs
febo: Yeah, this was me dealing with `RpcError(DeadlineExceeded)`. 😅

program/src/processor/withdraw_inactive_stake.rs
buffalojoec: Looks like the name of this field threw me off. How about we rename it to something like `vault_authority_bump`?

febo: Sounds good, makes it clearer what the bump is.

program/src/state/mod.rs
febo: The bump is stored on the account so we can save CU.

buffalojoec (commented):
This looks ready to go, but could we possibly get some test coverage on the vault and mint checks in the processor? Seems like the stake, config, and amount checks are well-covered, but the key-pointers would be good to cover.

program/src/instruction.rs
```diff
@@ -204,18 +204,23 @@ pub enum StakeInstruction {
     )]
     #[account(
         4,
+        name = "mint",
+        desc = "Stake Token Mint"
```
joncinque: nit: not a big deal at all, but I typically like to follow the account list as in the token program, so source, mint, destination, then authority

buffalojoec (approved):
Thanks for adding those tests!

program/src/processor/withdraw_inactive_stake.rs
```diff
@@ -17,10 +26,150 @@ use crate::instruction::accounts::{Context, WithdrawInactiveStakeAccounts};
...
+
+    stake.inactive_amount = stake
+        .inactive_amount
+        .checked_sub(amount)
+        .ok_or(ProgramError::ArithmeticOverflow)?;
+
+    config.token_amount_delegated = config
+        .token_amount_delegated
+        .checked_sub(amount)
+        .ok_or(ProgramError::ArithmeticOverflow)?;
```
joncinque: Maybe this is clarified later, but `amount` should be removed from `token_amount_delegated` during `InactivateStake`, and not during withdrawal

program/src/processor/withdraw_inactive_stake.rs
```diff
@@ -17,10 +26,150 @@ use crate::instruction::accounts::{Context, WithdrawInactiveStakeAccounts};
...
+    // Transfer the tokens from the vault to destination (unstakes them).
+
+    drop(vault_data);
+    drop(mint_data);
+
+    spl_token_2022::onchain::invoke_transfer_checked(
+        &spl_token_2022::ID,
+        ctx.accounts.vault.clone(),
+        ctx.accounts.mint.clone(),
+        ctx.accounts.destination_token_account.clone(),
```
joncinque, line 169: What do you think about adding a check to make sure that `vault != destination_token_account`? Not so much to protect users, but to avoid messing things up in the vault account

joncinque (commented):
Looks good! Just a few little things

program/src/processor/withdraw_inactive_stake.rs
febo: Makes sense, but this has an implication on the `Slash` instruction ability to slash inactive stake. I think we have a couple of options:
1. In the current design, we can slash inactive stake since it is still in the system (the `token_amount_delegated` "knows" about them). Spoke to Joe earlier, and one way to keep this capability is to create a `token_amount_inactive` on the `Config` account to track inactive stake which has not been withdrawn.
2. We don't slash inactive stake, so we modify the `Slash` instruction to only go through `active` and `deactivating` stake.

program/src/processor/withdraw_inactive_stake.rs
febo: Yeah, sounds good!

program/src/processor/withdraw_inactive_stake.rs
joncinque: I would go with number 2.

Imagine if you have staked SOL:

* you deactivate it
* you wait, it becomes inactive, but you don't withdraw
* the validator you were staked to commits a slashable offense
* you get slashed

That would be *very* bad

febo: Is there a risk that someone keeps track of the governance vote and as soon as it sees that the validator could be slashed, inactivate its stake? Not sure if there are guarantees that the voting period will always be shorter than the cooldown period.

I am thinking the scenario where:
* validator commits a slashable offence
* slash proposal begin voting phase
* every staker inactive their stake
* slash proposal is executed
* there are no tokens left to slash

Slashing inactive tokens does not prevent the scenario above, but adds an extra step.

joncinque: There's a 30-day cooldown period for deactivation, and it's only possible to deactivate 10% at a time, so even if the voting period takes 30 days (which shouldn't happen), they'll still lose 90% of their stake

febo: Cool, sounds good. I will update the logic.

joncinque (approved):
Looks great, thanks!

PR #25, opened 2024-07-11
Processor: `Slash`

febo:
This PR adds the processor for the`Slash` instruction.


buffalojoec:
Looks like maybe this one needs a rebase? Then happy to review.

febo:
> Looks like maybe this one needs a rebase? Then happy to review.

Rebased.

program/src/processor/slash.rs
```diff
@@ -11,14 +24,187 @@ use crate::instruction::accounts::{Context, SlashAccounts};
...
+    // slash authority
+    // - must be a signer
+    // - must match the slash authority on the config account
+    //
+    // When there is no slash authority set, the stake account cannot be slashed and
+    // an error is returned.
+
+    let slash_authority = config.slash_authority.into();
+
+    if let Some(slash_authority) = slash_authority {
```
buffalojoec: Up to you, but I think you can one-line this.

```suggestion
    if let Some(slash_authority) = Option::<Pubkey>::from(config.slash_authority) {
```

program/src/processor/slash.rs
```diff
@@ -11,14 +24,187 @@ use crate::instruction::accounts::{Context, SlashAccounts};
...
+        "mint"
+    );
+
+    // Update the stake amount on both stake and config accounts.
+
+    require!(
+        stake.amount >= amount,
+        StakeError::InsufficientStakeAmount,
+        "stake"
+    );
```
buffalojoec: I think the slash amount should be `min(stake.amount, amount)`, so that if they don't have enough in their stake account for the slash amount, they instead get slashed to zero.

program/src/processor/slash.rs
```diff
@@ -11,14 +24,187 @@ use crate::instruction::accounts::{Context, SlashAccounts};
...
+    stake.amount = stake
+        .amount
+        .checked_sub(amount)
+        .ok_or(ProgramError::ArithmeticOverflow)?;
+
+    require!(
+        config.token_amount_delegated >= amount,
+        StakeError::InsufficientStakeAmount,
+        "config"
+    );
```
buffalojoec: Same here I guess, but this shouldn't really ever happen.

program/src/processor/slash.rs
```diff
@@ -11,14 +24,187 @@ use crate::instruction::accounts::{Context, SlashAccounts};
...
+        .checked_sub(amount)
+        .ok_or(ProgramError::ArithmeticOverflow)?;
+
+    // Validates the amount of tokens to burn.
+
+    require!(
+        amount > 0,
+        StakeError::InvalidAmount,
+        "amount must be greater than 0"
+    );
```
buffalojoec: Should this be an error? Or should this check come earlier?

clients/rust/tests/slash.rs
```diff
@@ -0,0 +1,673 @@
...
+
+    // And the slhed stake account has no tokens.
+
+    let account = get_account!(context, stake_manager.stake);
+    let stake = Stake::from_bytes(account.data.as_ref()).unwrap();
+    assert_eq!(stake.amount, 0);
+}
+
+#[tokio::test]
+async fn fail_slash_with_zero_amount() {
```
buffalojoec, line 116: If we do decide to go with `min(stake.amount, amount)`, let's add a test for the case where `stake.amount < amount`.

program/src/processor/slash.rs
```diff
@@ -11,14 +24,187 @@ use crate::instruction::accounts::{Context, SlashAccounts};
...
+    // mint
+    // - must match the stake vault mint
+
+    require!(
+        &vault.base.mint == ctx.accounts.mint.key,
+        StakeError::InvalidMint,
+        "mint"
+    );
+
+    // Update the stake amount on both stake and config accounts.
```
buffalojoec: Should we also be trying to slash inactive stake? What if I have 50 SOL staked, but 100 inactive? Or better yet, I had 150 SOL staked and then inactivated 100 of it, because I knew I was going to be slashed. If the slash is 100 SOL, we could still execute the full slash successfully on both staked amount and inactive. What do you think?

buffalojoec (commented):
Nice, just have some general questions about how slashing should work.

program/src/processor/slash.rs
febo: Moved to the start of the amount validation.

program/src/processor/slash.rs
febo: Nice!

program/src/processor/slash.rs
febo: Agreed. The logic was updated to first try to slash the active amount; if there is not enough, slash inactive then deactivating; if still not enough, log a message but the instruction succeeds.

program/src/processor/slash.rs
febo: I left this one generating an error, since this seems critical. As you mentioned, this should never happen. What do you think?

clients/rust/tests/slash.rs
```diff
@@ -671,3 +670,166 @@ async fn fail_slash_with_wrong_config_account() {
...
+    )
+    .start_with_context()
+    .await;
+
+    // Given a config and stake accounts.
+
+    let config_manager = ConfigManager::new(&mut context).await;
+    let stake_manager = StakeManager::new(&mut context, &config_manager.config).await;
+
+    // And we set 50 tokens to the vault account.
```
buffalojoec, line 691: This is testing the case where the vault doesn't have enough tokens for slashing. Can we add one where the staker doesn't have enough tokens across active and inactive, and we observe they were slashed to zero but the vault was only deducted the amount the staker had?

program/src/processor/slash.rs
```diff
@@ -147,64 +150,100 @@ pub fn process_slash(
...
+
     stake.amount = stake
         .amount
-        .checked_sub(amount)
+        .checked_sub(slash_amount)
         .ok_or(ProgramError::ArithmeticOverflow)?;
 
+    let remaining = amount.saturating_sub(slash_amount);
+
+    if remaining > 0 && stake.inactive_amount > 0 {
```
buffalojoec: This confused me a bit at first. Can we do a little refactoring to make this a bit easier to follow?

```rust
let mut remaining_slash_amount = amount;

// First slash active stake.
let active_stake_to_slash = min(stake.amount, remaining_slash_amount);
stake.amount = stake
    .amount
    .checked_sub(active_stake_to_slash)
    .ok_or(ProgramError::ArithmeticOverflow)?;
remaining_slash_amount = remaining_slash_amount.saturating_sub(active_stake_to_slash);

// Then, if the slash amount has not been fulfilled, slash inactive stake.
if remaining_slash_amount > 0 {
    let inactive_stake_to_slash = min(stake.inactive_amount, remaining_slash_amount);
    stake.amount = stake
        .inactive_amount
        .checked_sub(inactive_stake_to_slash)
        .ok_or(ProgramError::ArithmeticOverflow)?;
    remaining_slash_amount = remaining_slash_amount.saturating_sub(inactive_stake_to_slash);
}

// Then, if the slash amount has not been fulfilled, slash deactivating stake.
// ...
```

It's pretty similar to what you have, just a bit more clear.

program/src/processor/slash.rs
```diff
@@ -147,64 +150,100 @@ pub fn process_slash(
...
-    // Validates the amount of tokens to burn.
+    if amount > slash_amount {
+        // The amount to slash was greater than the amount available on the
+        // stake account (active and inactive).
+        msg!(
+            "Slash amount greater than available tokens on stake account ({} required, {} available)",
+            amount,
+            slash_amount,
+        );
+    }
```
buffalojoec: re: my previous comment, then here you can just do something like:

```rust
let amount_to_burn = amount - remaining_slash_amount;
```

program/src/processor/slash.rs
buffalojoec: Yeah, an error definitely makes sense. I wonder if this error is confusing though, since it's designed for throwing when the staker's amount is too little. Should we introduce a new one for this case?

febo: Good point!

clients/rust/tests/slash.rs
febo: Done!

program/src/processor/slash.rs
febo: Nice, looks better now.

buffalojoec (approved):
Looks great!

program/src/processor/slash.rs
```diff
@@ -11,14 +27,257 @@ use crate::instruction::accounts::{Context, SlashAccounts};
...
+
+        stake.inactive_amount = stake
+            .inactive_amount
+            .checked_sub(inactive_stake_to_slash)
+            .ok_or(ProgramError::ArithmeticOverflow)?;
+        remaining_slash_amount = remaining_slash_amount.saturating_sub(inactive_stake_to_slash);
+    }
+
+    // Then, if the slash amount has not been fulfilled, slash deactivating stake
+    // (if there are deactivating tokens).
```
joncinque: The deactivating token amount should be a portion of the currently active tokens, and not a separate amount. So if you've already slashed all of the active amount, that would also take the deactivating amount.

In fact, there should be an additional bit to make sure that deactivating amount isn't greater than amount after the slash, and should be set to `min(stake.deactivating_amount, stake.amount)`

program/src/processor/slash.rs
```diff
@@ -11,14 +27,257 @@ use crate::instruction::accounts::{Context, SlashAccounts};
...
+    // First slash active stake.
+    let active_stake_to_slash = min(amount, stake.amount);
+    stake.amount = stake
+        .amount
+        .checked_sub(active_stake_to_slash)
+        .ok_or(ProgramError::ArithmeticOverflow)?;
+    remaining_slash_amount = remaining_slash_amount.saturating_sub(active_stake_to_slash);
+
+    // Then, if the slash amount has not been fulfilled, slash inactive stake
+    // (if there are inactive tokens).
```
joncinque: Normally, inactive tokens shouldn't be touchable by slashing -- the whole point is that they're inactive! But we can double-check with Uri about this to be totally sure.

program/src/processor/slash.rs
```diff
@@ -11,14 +27,257 @@ use crate::instruction::accounts::{Context, SlashAccounts};
...
+            // if there are no more deactivating tokens, clear the
+            // deactivation timestamp
+            stake.deactivation_timestamp = None;
+        }
+
+        remaining_slash_amount = remaining_slash_amount.saturating_sub(deactivating_stake_to_slash);
+    }
+
+    // Finally, the adjusted amount to slash.
+    let amount_to_slash = amount - remaining_slash_amount;
```
joncinque: Going with the other comments, this should just be `active_stake_to_slash`

joncinque (commented):
Just a few comments, mainly on how the values should be interpreted

program/src/processor/slash.rs
febo: True, good point, makes no sense since the `amount` only gets updated when they move from `deactivating` to `inactive`. I will fix it.

program/src/processor/slash.rs
```diff
@@ -11,14 +27,219 @@ use crate::instruction::accounts::{Context, SlashAccounts};
...
+        .amount
+        .checked_sub(active_stake_to_slash)
+        .ok_or(ProgramError::ArithmeticOverflow)?;
+
+    if stake.deactivating_amount > stake.amount {
+        // adjust the deactivating amount if it's greater than the "active" stake
+        // after slashing
+        stake.deactivating_amount = stake.amount;
+        // clear the deactivation timestamp if the stake amount is 0
+        if stake.amount == 0 {
```
joncinque: nit: sorry, this check made me second-guess quite a bit -- how about checking if `stake.deactivating_amount == 0` to make it clearer?

program/src/processor/slash.rs
```diff
@@ -11,14 +27,219 @@ use crate::instruction::accounts::{Context, SlashAccounts};
...
+        active_stake_to_slash,
+        config.token_amount_delegated
+    );
+
+    config.token_amount_delegated = config
+        .token_amount_delegated
+        .checked_sub(active_stake_to_slash)
+        .ok_or(ProgramError::ArithmeticOverflow)?;
+
+    if amount.saturating_sub(active_stake_to_slash) > 0 {
```
joncinque: nit: maybe this is just me, but I find `amount > active_stake_to_slash` to be clearer than saturating math

clients/rust/tests/slash.rs
```diff
@@ -0,0 +1,938 @@
...
+
+    let tx = Transaction::new_signed_with_payer(
+        &[slash_ix],
+        Some(&context.payer.pubkey()),
+        &[&context.payer, &config_manager.authority],
+        context.last_blockhash,
+    );
+    context.banks_client.process_transaction(tx).await.unwrap();
+
+    // Then the slash is successful but not token is burned.
```
joncinque: micro-nit:
```suggestion
    // Then the slash is successful but no tokens are burned.
```

joncinque (commented):
Just a few last nits, then this is good to go!

program/src/processor/slash.rs
febo: Definitely clearer!

joncinque (approved):
Thanks!

PR #26, opened 2024-07-12
Processor: `HarvestHolderRewards`

febo:
This PR adds the processor for the `HarvestHolderRewards` instruction.


joncinque:
This seems to add quite a bit more than just "harvest" -- does it need to be rebased?

febo:
> This seems to add quite a bit more than just "harvest" -- does it need to be rebased?

Yeah, it was not a good idea to have merged other PRs on top of this one - sorry about that. I think it would be easier to look at the commit list:

https://github.com/paladin-bladesmith/stake-program/pull/26/commits

and ignore the entry for #27 (Processor: HarvestStakeRewards).

joncinque:
Since we'll have to merge this PR at some point, which will require cleaning up the history, I'd prefer to have the commit history sorted out before looking at it. Would that be possible?

febo:
Yep, I will re-create the PR.

PR #27, opened 2024-07-13
Processor: `HarvestStakeRewards`

febo:
This PR adds the processor for the `HarvestStakeRewards` instruction.

### DIFF
--- a/clients/rust/tests/harvest_stake_rewards.rs
+++ b/clients/rust/tests/harvest_stake_rewards.rs
@@ -49,7 +49,7 @@ async fn harvest_stake_rewards() {
     // "manually" set the total amount delegated
     config_account.token_amount_delegated = 100;
     config_account.accumulated_stake_rewards_per_token =
-        calculate_stake_rewards_per_token(4_000_000_000, 100).to_le_bytes();
+        calculate_stake_rewards_per_token(4_000_000_000, 100);
 
     account.lamports += 4_000_000_000;
     account.data = config_account.try_to_vec().unwrap();
@@ -187,7 +187,7 @@ async fn harvest_stake_rewards_after_harvesting() {
     // "manually" set the total amount delegated
     config_account.token_amount_delegated = 100;
     config_account.accumulated_stake_rewards_per_token =
-        calculate_stake_rewards_per_token(4_000_000_000, 100).to_le_bytes();
+        calculate_stake_rewards_per_token(4_000_000_000, 100);
 
     account.lamports += 4_000_000_000;
     account.data = config_account.try_to_vec().unwrap();
@@ -301,7 +301,7 @@ async fn fail_harvest_stake_rewards_with_wrong_authority() {
     // "manually" set the total amount delegated
     config_account.token_amount_delegated = 100;
     config_account.accumulated_stake_rewards_per_token =
-        calculate_stake_rewards_per_token(4_000_000_000, 100).to_le_bytes();
+        calculate_stake_rewards_per_token(4_000_000_000, 100);
 
     account.lamports += 4_000_000_000;
     account.data = config_account.try_to_vec().unwrap();
@@ -432,7 +432,7 @@ async fn fail_harvest_stake_rewards_with_uninitialized_stake_account() {
     // "manually" set the total amount delegated
     config_account.token_amount_delegated = 100;
     config_account.accumulated_stake_rewards_per_token =
-        calculate_stake_rewards_per_token(4_000_000_000, 100).to_le_bytes();
+        calculate_stake_rewards_per_token(4_000_000_000, 100);
 
     account.lamports += 4_000_000_000;
     account.data = config_account.try_to_vec().unwrap();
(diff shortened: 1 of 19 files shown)


program/src/state/config.rs
```diff
@@ -37,6 +39,12 @@ pub struct Config {
     /// Running total of all stake rewards distributed.
     pub total_stake_rewards: u64,
```
joncinque: I might be missing something, but I think we can remove this since we have `accumulated_stake_rewards_per_token`

program/src/state/config.rs
```diff
@@ -45,12 +53,35 @@ pub struct Config {
...
 }
 
 impl Config {
     pub const LEN: usize = std::mem::size_of::<Config>();
 
+    pub fn new(
+        authority: OptionalNonZeroPubkey,
+        slash_authority: OptionalNonZeroPubkey,
+        vault: Pubkey,
+        cooldown_time_seconds: i64,
```
joncinque: nit: let's have this take a `u64` and do the conversion to make it harder to use incorrectly, since a negative value is wrong

program/src/processor/harvest_stake_rewards.rs
```diff
@@ -13,8 +21,123 @@ use crate::instruction::accounts::{Context, HarvestStakeRewardsAccounts};
...
+    // Determine the stake rewards.
+
+    let accumulated_rewards_per_token = config.accumulated_stake_rewards_per_token();
+    let rewards = calculate_eligible_rewards(
+        accumulated_rewards_per_token,
+        stake.last_seen_stake_rewards_per_token(),
+        stake.amount,
+    )?;
+    // update the last seen holder rewards
+    stake.set_last_seen_stake_rewards_per_token(accumulated_rewards_per_token);
```
joncinque: Let's only update this if rewards != 0, otherwise people can lose out on fractional rewards. And can we print a message if rewards == 0 so people know why they didn't get anything?

program/src/processor/harvest_stake_rewards.rs
```diff
@@ -13,8 +21,123 @@ use crate::instruction::accounts::{Context, HarvestStakeRewardsAccounts};
...
+            .ok_or(ProgramError::ArithmeticOverflow)?;
+        let updated_destination_lamports = ctx
+            .accounts
+            .destination
+            .lamports()
+            .checked_add(rewards)
+            .ok_or(ProgramError::ArithmeticOverflow)?;
+
+        **ctx.accounts.config.try_borrow_mut_lamports()? = updated_config_lamports;
+        **ctx.accounts.destination.try_borrow_mut_lamports()? = updated_destination_lamports;
```
joncinque, line 140: not-important-but-something-to-think-about: I wonder if it's worth having some defensive checks on `destination`. For example, making sure that it isn't the config or the stake account. I eventually settled that we don't need them, but I could be wrong.

clients/rust/tests/harvest_stake_rewards.rs
```diff
@@ -0,0 +1,551 @@
...
+    context.banks_client.process_transaction(tx).await.unwrap();
+
+    // Then the transaction succeeds but the destination account has no rewards.
+
+    let account = context.banks_client.get_account(destination).await.unwrap();
+    assert!(account.is_none());
+}
+
+#[tokio::test]
+async fn harvest_holder_rewards_after_harvesting() {
```
joncinque: nit: this whole test uses "holder" rewards in a few places, but it's all about stake rewards
```suggestion
async fn harvest_stake_rewards_after_harvesting() {
```

clients/rust/tests/harvest_stake_rewards.rs
```diff
@@ -0,0 +1,551 @@
...
+        .await
+        .unwrap_err();
+
+    // Then we expect an error.
+
+    assert_instruction_error!(err, InstructionError::UninitializedAccount);
+}
+
+#[tokio::test]
+async fn fail_harvest_holder_rewards_with_uninitialized_stake_account() {
```
joncinque: nit:
```suggestion
async fn fail_harvest_stake_rewards_with_uninitialized_stake_account() {
```

joncinque (commented):
Looks great! Just a few small things

program/src/state/config.rs
febo: Good point, yes, it seems that we can remove it.

program/src/processor/harvest_stake_rewards.rs
febo: We could add a check against `config` and `stake` to avoid mistakes, but it would be difficult to cover all cases. I first thought of restricting to accounts owned by system program, but there might be cases that you want to send the rewards to a PDA (e.g., a multisig account).

clients/rust/tests/harvest_stake_rewards.rs
febo: Oh, my bad.

program/src/processor/harvest_stake_rewards.rs
joncinque: That makes sense, let's leave out any checks then

joncinque (approved):

PR #28, opened 2024-07-15
Processor: `DistributeRewards`

febo:
This PR adds the processor for the `DistributeRewards` instruction.

### DIFF
--- a/program/src/processor/distribute_rewards.rs
+++ b/program/src/processor/distribute_rewards.rs
@@ -62,9 +62,6 @@ pub fn process_distribute_rewards(
     );
 
     // Calculates the rewards per token
-    //
-    // TODO: Do we still transfer the rewards to the config account if there are no
-    // staked tokens?
 
     let rewards_per_token =
         calculate_stake_rewards_per_token(amount, config.token_amount_delegated)?;
@@ -76,16 +73,16 @@ pub fn process_distribute_rewards(
             .checked_add(rewards_per_token)
             .ok_or(ProgramError::ArithmeticOverflow)?;
         config.set_accumulated_stake_rewards_per_token(accumulated);
-
-        // Transfers the rewards to the config account.
-
-        drop(config_data);
-
-        invoke(
-            &system_instruction::transfer(ctx.accounts.payer.key, ctx.accounts.config.key, amount),
-            &[ctx.accounts.payer.clone(), ctx.accounts.config.clone()],
-        )?;
     }
+
+    // Transfers the rewards to the config account.
+
+    drop(config_data);
+
+    invoke(
+        &system_instruction::transfer(ctx.accounts.payer.key, ctx.accounts.config.key, amount),
+        &[ctx.accounts.payer.clone(), ctx.accounts.config.clone()],
+    )?;
 
     Ok(())
 }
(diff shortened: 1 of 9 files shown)


scripts/generate-clients.mjs
joncinque, line 1: nit: Most of these changes are cosmetic -- were they meant to be included in this PR?

program/src/processor/distribute_rewards.rs
```diff
@@ -10,9 +17,75 @@ use crate::instruction::accounts::{Context, DistributeRewardsAccounts};
...
+    require!(
+        ctx.accounts.system_program.key == &system_program::ID,
+        ProgramError::IncorrectProgramId,
+        "system program"
+    );
+
+    // Calculates the rewards per token
+    //
+    // TODO: Do we still transfer the rewards to the config account if there are no
+    // staked tokens?
```
joncinque: I think you've done the right behavior of silently doing nothing if there are no staked tokens. It would be useful to log a message saying that the tokens were not transferred because nothing is staked.

joncinque (approved):
Looks good! Just tiny things that don't stop this from going in as is. You might just want to remove the TODO

program/src/processor/distribute_rewards.rs
buffalojoec: Unfortunately I don't think this is the right thing to do here.

If you have no staked tokens, then don't update the `stake_rewards_per_token` rate (since it's 0 as a result of the 0 denominator), but still move the lamports. What if I want to feed rewards to the stake config while the staking ecosystem is still setting up, or not set up yet at all?

The rewards program doesn't prohibit the movement of reward lamports if the token supply is zero, it just doesn't update the rate.

https://github.com/paladin-bladesmith/rewards-program/blob/26845c684b485d7ea90c3332793afcd8316ef58b/program/src/processor.rs#L363-L385

joncinque: I was leaning towards that at first. If we do that, however, the pool is in a position where there are untouchable rewards. At which point, we, as the developers, need to change the program to make those rewards available. Rather than put ourselves in a position where we need to hack the program later, it's much easier for nothing to happen or for the program to throw an error, so that the rewards can be distributed later when something is staked.

buffalojoec: How does this affect the top-level system reward distribution then? If I need to send 10% to the governance treasury, 40% to stakers, and 50% to holders, and there are no staked tokens, then what happens to that 40%?

joncinque: Ah, that's an excellent point. Let's go with your suggestion then and transfer the tokens no matter what to be consistent

buffalojoec: Fwiw, the rewards only become untouchable when the system has no tokens, which would either be when it first launches or if everyone pulls their stake. In either case, I'd say the reward issuer should be aware of this and probably stop the funnel at the top.

buffalojoec: Same idea for distributing holder rewards with no tokens. DAO/Management concern.

scripts/generate-clients.mjs
febo: Not intentionally, but I updated my prettier config on vscode so it now applies to all typescript files.

The only "relevant" change is the rename of the instruction args: https://github.com/paladin-bladesmith/stake-program/blob/bb67b84fe32c46a30544636739fc82d07f3cb5ee/scripts/generate-clients.mjs#L141-L152

program/src/processor/distribute_rewards.rs
febo: Ok, changed to transfer the rewards but not update the `accumulated_stake_rewards_per_token` rate.

joncinque (approved):

PR #29, opened 2024-07-16
rename crate and set version

buffalojoec:


program/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
-name = "stake-program"
-version = "0.0.0"
+name = "paladin-stake-program"
+version = "0.0.1"
 edition = "2021"
 readme = "./README.md"
 license-file = "../LICENSE"
```
febo, line 6: If the plan is to publish this crate, then we need to remove the line below:
```
publish = false
```

clients/js/src/generated/index.ts
```diff
@@ -9,6 +9,5 @@
 export * from './accounts';
 export * from './errors';
 export * from './instructions';
-export * from './pdas';
```
febo: I can't comment on the `generate-clients.mjs` file, but we are missing a rename there. This is why the `pdas` folder is not being generated.

https://github.com/paladin-bladesmith/stake-program/blob/62b2dc6dce04b520e411d2e2fa82907eacdb0b76/scripts/generate-clients.mjs#L26

```
"[programNode]stake" -> "[programNode]paladinStakeProgram"

febo:
We can also remove lines `15-20` from the Kinobi configuration since we are not renaming the program anymore.

https://github.com/paladin-bladesmith/stake-program/blob/62b2dc6dce04b520e411d2e2fa82907eacdb0b76/scripts/generate-clients.mjs#L15-L20

buffalojoec:
Thanks @febo, all set!

febo (approved):
Looks great!

PR #30, opened 2024-07-17
Update stake account derivation

febo:
This PR updates the `Stake` account derivation to use the `validator vote` address instead of the `node id`.

### DIFF
--- a/program/src/state/stake.rs
+++ b/program/src/state/stake.rs
@@ -37,7 +37,7 @@ pub struct Stake {
     pub authority: Pubkey,
 
     /// The address of the validator vote account
-    pub validator: Pubkey,
+    pub validator_vote: Pubkey,
 
     /// Stores the "last_seen_holder_rewards" just for this stake account, allowing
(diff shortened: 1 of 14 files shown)


program/src/state/stake.rs
```diff
@@ -56,15 +56,15 @@ impl Stake {
...
+    pub fn new(authority: Pubkey, validator_vote: Pubkey) -> Self {
         Self {
             discriminator: Stake::SPL_DISCRIMINATOR.into(),
             amount: u64::default(),
             deactivation_timestamp: Option::default(),
             deactivating_amount: u64::default(),
             inactive_amount: u64::default(),
             authority,
-            validator,
+            validator: validator_vote,
```
joncinque: nit: can we have the field name be `validator_vote` for consistency and clarity?

joncinque (approved):
Just one naming nit, but the rest looks good!

program/src/state/stake.rs
febo: Done!

joncinque (approved):

PR #31, opened 2024-07-17
Update deactivate stake logic

febo:
This PR updates the logic on the `DeactivateStake`:
- Cancel deactivation when `amount` is `0`
- Use `u128` in the deactivation amount calculation
- Use the `Stake` amount to calculate the deactivation ratio

It also adds validation for the `max_deactivation_basis_points` value on `InitializeConfig` and `UpdateConfig`.


program/src/processor/deactivate_stake.rs
```diff
@@ -12,13 +12,16 @@ use crate::{
     state::{find_stake_pda, Config, Stake},
 };
 
+const MAX_BASIS_POINTS: u128 = 10_000;
```
joncinque: side note: unless I'm missing something, there's no check during initialize config or update config that the `basis_points <= 10_000`. we should add that in

joncinque (approved):
Looks great, thanks! Noticed another thing to fix :sweat_smile:

buffalojoec (approved):

program/src/processor/deactivate_stake.rs
febo: Oh, good catch, definitely need to add that.

febo: Done.

program/src/processor/deactivate_stake.rs
```diff
@@ -107,14 +108,10 @@ pub fn process_deactivate_stake(
...
-    require!(amount > 0, StakeError::InvalidAmount);
-
     require!(stake.amount >= amount, StakeError::InsufficientStakeAmount);
 
-    let max_deactivation_amount = get_max_deactivation_amount(
-        config.token_amount_delegated,
-        config.max_deactivation_basis_points,
-    )?;
+    let max_deactivation_amount =
+        get_max_deactivation_amount(stake.amount, config.max_deactivation_basis_points)?;
```
joncinque, line 114: oops, I didn't even realize this was incorrect before, nice fix!

joncinque (approved):

program/src/processor/deactivate_stake.rs
febo: Yeah, it was something that we picked up with Joe when we went through the docs again last week. 😀

PR #33, opened 2024-07-25
Processor: `HarvestHolderRewards`

febo:
This PR adds the processor for the `HarvestHolderRewards` instruction.

### DIFF
--- a/clients/rust/tests/harvest_holder_rewards.rs
+++ b/clients/rust/tests/harvest_holder_rewards.rs
@@ -1,0 +1,783 @@
+#![cfg(feature = "test-sbf")]
+
+mod setup;
+
+use borsh::BorshSerialize;
+use paladin_rewards_program_client::accounts::HolderRewards;
+use paladin_stake_program_client::{
+    accounts::{Config, Stake},
+    errors::PaladinStakeProgramError,
+    instructions::HarvestHolderRewardsBuilder,
+    pdas::{find_stake_pda, find_vault_pda},
+};
+use setup::{
+    config::ConfigManager,
+    rewards::{create_holder_rewards, create_holder_rewards_pool},
+    stake::StakeManager,
+};
+use solana_program_test::{tokio, ProgramTest};
+use solana_sdk::{
+    account::{Account, AccountSharedData},
+    instruction::InstructionError,
+    pubkey::Pubkey,
+    signature::{Keypair, Signer},
+    transaction::Transaction,
+};
+
+#[tokio::test]
+async fn harvest_holder_rewards() {
+    let mut program_test = ProgramTest::new(
+        "paladin_stake_program",
+        paladin_stake_program_client::ID,
+        None,
+    );
+    program_test.add_program(
+        "paladin_rewards_program",
+        paladin_rewards_program_client::ID,
+        None,
+    );
+    let mut context = program_test.start_with_context().await;
+
+    // Given a config account with 100 staked amount.
+
+    let config_manager = ConfigManager::new(&mut context).await;
+
+    let mut account = get_account!(context, config_manager.config);
+    let mut config_account = Config::from_bytes(account.data.as_ref()).unwrap();
+    // "manually" set the total amount delegated
+    config_account.token_amount_delegated = 100;
+    account.data = config_account.try_to_vec().unwrap();
+    context.set_account(&config_manager.config, &account.into());
+
+    // And 4 SOL rewards on its vault account.
+
+    let mut account = get_account!(context, config_manager.vault);
+    account.lamports = account.lamports.saturating_add(4_000_000_000);
+    let rewards_lamports = account.lamports;
+    context.set_account(&config_manager.vault, &account.into());
+
+    // And a stake account wiht a 50 staked amount.
+
+    let stake_manager = StakeManager::new(&mut context, &config_manager.config).await;
+
+    let mut account = get_account!(context, stake_manager.stake);
+    let mut stake_account = Stake::from_bytes(account.data.as_ref()).unwrap();
+    // "manually" set the amount to 50
+    stake_account.amount = 50;
+
+    account.data = stake_account.try_to_vec().unwrap();
+    context.set_account(&stake_manager.stake, &account.into());
+
+    // And we initialize the holder rewards accounts.
+
+    let holder_rewards_pool = create_holder_rewards_pool(
+        &mut context,
+        &config_manager.mint,
+        &config_manager.mint_authority,
+    )
+    .await;
+
+    let holder_rewards = create_holder_rewards(
+        &mut context,
+        &holder_rewards_pool,
+        &config_manager.mint,
+        &config_manager.vault,
+    )
+    .await;
+
+    let mut account = get_account!(context, holder_rewards);
+    let mut holder_rewards_account = HolderRewards::from_bytes(account.data.as_ref()).unwrap();
+    // "manually" set last accumulated rewards per token to 40_000_000 (0.04 SOL)
+    holder_rewards_account.last_accumulated_rewards_per_token = 40_000_000 * 1_000_000_000;
+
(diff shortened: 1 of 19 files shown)


program/src/processor/harvest_holder_rewards.rs
```diff
@@ -21,11 +36,231 @@ use crate::instruction::accounts::{Context, HarvestHolderRewardsAccounts};
...
+    require!(
+        ctx.accounts.holder_rewards.owner == &paladin_rewards_program_client::ID,
+        ProgramError::InvalidAccountOwner,
+        "holder rewards",
+    );
+
+    let (derivation, _) = Pubkey::find_program_address(
+        &["holder".as_bytes(), ctx.accounts.vault.key.as_ref()],
+        &paladin_rewards_program_client::ID,
+    );
```
buffalojoec: I _swore_ we added helpers for this, what happened to those? I can't find them in the Rust client anymore.

program/src/state/mod.rs
```diff
@@ -57,3 +66,31 @@ pub fn get_stake_pda_signer_seeds<'a>(
         bump_seed,
     ]
 }
+
+/// Calculate the eligible rewards for a given token account balance.
+///
+/// This uses the same logic as the Rewards program.
+pub fn calculate_eligible_rewards(
```
buffalojoec, line 73: Lmk if you think we should somehow land this in the Rewards program client so we don't have to reimplement here.

program/src/processor/harvest_holder_rewards.rs
```diff
@@ -21,11 +36,231 @@ use crate::instruction::accounts::{Context, HarvestHolderRewardsAccounts};
...
+    if rewards != 0 {
+        // update the last seen holder rewards
+        stake.set_last_seen_holder_rewards_per_token(
+            holder_rewards.last_accumulated_rewards_per_token,
+        );
+
+        // Rewards are stored on the `vault` token account. We need to first withdraw the excess lamports
+        // from the vault token account to the vault authority account. Then transfer the rewards amount
+        // to the destination account and send the remaining excess lamports back to the vault token
+        // account.
```
buffalojoec, line 195: Wouldn't it be easier to withdraw excess lamports to the destination account, then send the remaining excess lamports back to the vault token account? That would save a jump in between.

program/src/processor/harvest_holder_rewards.rs
```diff
@@ -21,11 +36,231 @@ use crate::instruction::accounts::{Context, HarvestHolderRewardsAccounts};
...
+            ctx.accounts.vault.key,
+            ctx.accounts.vault_authority.key,
+            ctx.accounts.vault_authority.key,
+            &[],
+        )?;
+
+        let signer_seeds = get_vault_pda_signer_seeds(ctx.accounts.config.key, &signer_bump);
+        // Stores the current lamports of the vault authority account before the withdraw so
+        // we can calculate the amount of rewards withdrawn.
+        let current_lamports = ctx.accounts.vault_authority.lamports();
```
buffalojoec: nit: any chance we can make this a bit more clear? It becomes hard to follow after a few steps below.

```suggestion
        let vault_authority_starting_lamports = ctx.accounts.vault_authority.lamports();
```

program/src/processor/harvest_holder_rewards.rs
```diff
@@ -21,11 +36,231 @@ use crate::instruction::accounts::{Context, HarvestHolderRewardsAccounts};
...
+        // If the withdraw does not result in enough lamports to cover the rewards, only
+        // harvest the available lamports. This should never happen, but the check is a
+        // failsafe.
+        let rewards = std::cmp::min(
+            rewards,
+            ctx.accounts
+                .vault_authority
+                .lamports()
+                .saturating_sub(current_lamports),
+        );
```
buffalojoec: What happens if this resolves to zero? In the case of no excess lamports on the vault, for example.

clients/rust/tests/harvest_holder_rewards.rs
```diff
@@ -0,0 +1,783 @@
...
+        context.last_blockhash,
+    );
+    context.banks_client.process_transaction(tx).await.unwrap();
+
+    // Then the destination account has the rewards.
+
+    let account = get_account!(context, destination);
+    assert_eq!(account.lamports, 2_000_000_000);
+
+    // And there should be rewards left on the vault account.
```
buffalojoec, line 133: Maybe just out of an abundance of caution, check that the vault authority account didn't hang onto any lingering lamports?

clients/rust/tests/harvest_holder_rewards.rs
```diff
@@ -0,0 +1,783 @@
...
+    );
+
+    // And the stake account last seen holder rewards per token is update.
+
+    let account = get_account!(context, stake_manager.stake);
+    let stake_account = Stake::from_bytes(account.data.as_ref()).unwrap();
+    assert_eq!(
+        stake_account.last_seen_holder_rewards_per_token,
+        40_000_000 * 1_000_000_000
+    );
```
buffalojoec, line 407: This gets set from the first harvest, though, right? The second one should see zero rewards eligible and terminate early?

buffalojoec (commented):
Looking great so far!!

program/src/processor/harvest_holder_rewards.rs
febo: It is there, just updated to use it.

program/src/processor/harvest_holder_rewards.rs
febo: The destination is not a signer, so I don't think we can do that.

program/src/processor/harvest_holder_rewards.rs
febo: It will invoke the `transfer` with `0`, which essentially will do nothing. In practice, this should not happen since we checked that we have rewards to distribute. Perhaps we can add a log message?

program/src/state/mod.rs
joncinque: +1 on this, can be done in a future PR though

clients/rust/tests/harvest_holder_rewards.rs
febo: Yes, the test is asserting that the value of `last_seen_holder_rewards_per_token` did not change from the first harvest.

program/src/state/mod.rs
febo: Yeah, might be a good idea.

program/src/processor/harvest_holder_rewards.rs
```diff
@@ -21,11 +36,228 @@ use crate::instruction::accounts::{Context, HarvestHolderRewardsAccounts};
...
+                ctx.accounts.vault_authority.key,
+                ctx.accounts.destination.key,
+                rewards,
+            ),
+            &[
+                ctx.accounts.vault_authority.clone(),
+                ctx.accounts.destination.clone(),
+            ],
+            &[&signer_seeds],
+        )?;
```
joncinque, line 244: I think it's OK either way, but we could add a check that `vault_authority != destination`

program/src/processor/harvest_holder_rewards.rs
joncinque: Yeah, since `withdraw_excess_lamports` takes everything, we unfortunately need to do this extra step

joncinque (approved):
Looks great to me!

program/src/processor/harvest_holder_rewards.rs
febo: Done - added a test for it as well.

program/src/processor/harvest_holder_rewards.rs
buffalojoec: > The destination is not a signer, so I don't think we can do that.

Ohhh right I see, you can't send from it.

buffalojoec (approved):

PR #34, opened 2024-07-26
Processor: `HarvestStakeRewards`

febo:
This PR adds the processor for the `HarvestStakeRewards` instruction.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -225,4 +225,4 @@ jobs:
           path: ./**/*.so
           key: ${{ runner.os }}-builds-${{ github.sha }}
       - name: Test Rust client
-        run: pnpm clients:rust:test
+        run: RUST_LOG=error pnpm clients:rust:test
--- a/clients/rust/tests/slash.rs
+++ b/clients/rust/tests/slash.rs
@@ -244,7 +244,7 @@ async fn slash_with_no_staked_amount() {
     );
     context.banks_client.process_transaction(tx).await.unwrap();
 
-    // Then the slash is successful but not token is burned.
+    // Then the slash is successful but no token is burned.
 
     let account = get_account!(context, config_manager.vault);
     let vault =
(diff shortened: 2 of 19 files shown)


.github/workflows/main.yml
```diff
@@ -225,4 +225,4 @@ jobs:
           path: ./**/*.so
           key: ${{ runner.os }}-builds-${{ github.sha }}
       - name: Test Rust client
-        run: pnpm clients:rust:test
+        run: RUST_LOG=error pnpm clients:rust:test
```
joncinque, line 228: nit: Was this intended?

clients/rust/tests/slash.rs
```diff
@@ -244,7 +244,7 @@ async fn slash_with_no_staked_amount() {
     );
     context.banks_client.process_transaction(tx).await.unwrap();
 
-    // Then the slash is successful but no token is burned.
+    // Then the slash is successful but not token is burned.
```
joncinque: not a big deal, but it looks like this picked up some older changes

program/src/processor/harvest_stake_rewards.rs
```diff
@@ -13,8 +21,127 @@ use crate::instruction::accounts::{Context, HarvestStakeRewardsAccounts};
...
+    );
+
+    // Determine the stake rewards.
+
+    let accumulated_rewards_per_token = config.accumulated_stake_rewards_per_token();
+    let rewards = calculate_eligible_rewards(
+        accumulated_rewards_per_token,
+        stake.last_seen_stake_rewards_per_token(),
+        stake.amount,
+    )?;
```
joncinque, line 102: Actually, considering this is using `calculate_eligible_rewards` too, we might want to keep a copy of the function in this repo rather than reusing the one from the rewards program.

joncinque (approved):
Looks great!

.github/workflows/main.yml
febo: Yes – this is part of my attempt to minimize the chance of `RpcError(DeadlineExceeded)`. Not sure if makes sense though. My rationale was to avoid "wasting" time writing the log.

program/src/processor/harvest_stake_rewards.rs
febo: @buffalojoec We might not need the `calculate_eligible_rewards` helper on the client.

PR #39, opened 2024-07-30
Refactor: Add `SolStakerStake` account

febo:
### Problem

The Stake program needs to support two types of stake accounts: one representing validators and another SOL stakers.

### Solution

Introduce a new stake account `SolStakerStake` to represent SOL stakers. The current stake account is renamed to `ValidatorStake`. Shared information between the accounts is stored in a common `Delegation` struct. Instructions that work with both account types are refactored.

Most of the tests are temporarily disabled, since stake creation will be refactored in subsequent PRs.


program/src/processor/distribute_rewards.rs
```diff
@@ -68,11 +69,10 @@ pub fn process_distribute_rewards(
 
     if rewards_per_token != 0 {
         // updates the accumulated stake rewards per token
-        let accumulated = config
-            .accumulated_stake_rewards_per_token()
+        let accumulated = <PodU128 as Into<u128>>::into(config.accumulated_stake_rewards_per_token)
```
joncinque: nit: I typically find it easier to read `from` instead of into in a situation like this, ie
```suggestion
        let accumulated = u128::from(config.accumulated_stake_rewards_per_token)
```

program/src/processor/initialize_stake.rs
```diff
@@ -8,7 +8,7 @@ use crate::{
     instruction::accounts::{Context, InitializeStakeAccounts},
     require,
     state::{
-        find_validator_stake_pda, get_validator_stake_pda_signer_seeds, Config, ValidatorStake,
+        find_validator_stake_pda, get_validator_stake_pda_signer_seeds, Config, SolStakerStake,
```
joncinque: This instruction is initializing a sol staker stake, but using the validator stake pda. I know that this is part of a string of refactors, but can we keep this consistent at least and revert the change in this file?

program/src/processor/mod.rs
```diff
@@ -144,3 +161,79 @@ macro_rules! require {
...
+
+        require!(
+            account.is_initialized(),
+            ProgramError::UninitializedAccount,
+            $name,
+        );
+
+        account
+    }};
+}
```
joncinque: nit: if we introduce an `IsInitialized` trait with an `is_initialized` function, and impl it with `ValidatorStake` and `SolStakerStake`, we can make this a function generic on `IsInitialized` instead of a macro

program/src/processor/mod.rs
```diff
@@ -144,3 +161,79 @@ macro_rules! require {
...
+///
+/// This function will validate that the account data is initialized and deriavation matches
+/// the expected PDA derivation.
+#[inline]
+pub fn unpack_delegation_mut(
+    stake_data: &mut [u8],
+    stake: &Pubkey,
+    config: &Pubkey,
+    program_id: &Pubkey,
+) -> Result<Delegation, ProgramError> {
```
joncinque: Maybe I'm missing something, but shouldn't this return a `&mut Delegation` to modify in place? Otherwise we need to serialize the changes back into the account

program/src/processor/mod.rs
```diff
@@ -144,3 +161,79 @@ macro_rules! require {
...
+
+    Ok(delegation)
+}
+
+/// Unpacks the delegation information from either a `SolStakerStake` and `ValidatorStake`
+/// accounts.
+///
+/// This function will validate that the account data is initialized.
+#[inline]
+pub fn unpack_delegation_mut_uncheked(stake_data: &mut [u8]) -> Result<Delegation, ProgramError> {
```
joncinque: Same with this, shouldn't it return a `&mut Delegation`?

program/src/processor/slash.rs
```diff
@@ -74,7 +74,7 @@ pub fn process_slash(
     );
 
     let mut stake_data = ctx.accounts.stake.try_borrow_mut_data()?;
-    let stake = bytemuck::try_from_bytes_mut::<ValidatorStake>(&mut stake_data)
+    let stake = bytemuck::try_from_bytes_mut::<SolStakerStake>(&mut stake_data)
```
joncinque: I know that it's disabled, but can you revert the changes in this file since we never want sol stakers to get slashed?

joncinque (commented):
Looks good overall! Just a few small things to sort out

program/src/processor/distribute_rewards.rs
febo: Nice, definitely better!

program/src/processor/initialize_stake.rs
febo: Done.

program/src/processor/mod.rs
febo: The changes are done in place since the `Delegation` is moved out of the stake struct, so the caller gets "full" ownership of it. We probably can't return a reference since the stake object will be out of the scope after the function returns.

program/src/processor/mod.rs
febo: Same as the comment above.

program/src/processor/mod.rs
joncinque: I don't think this is the case. I ran a little test on the sol stake view program, check it out: https://github.com/paladin-bladesmith/sol-stake-view-program/pull/4

program/src/processor/mod.rs
febo: Good suggestion, we are able to use `solana_program::program_pack::IsInitialized` and the code looks nicer now.

program/src/processor/mod.rs
febo: You are absolutely right! `Pod` types must implement `Copy`, so a copy is being returned. 🤦‍♂️

febo: ~We can't return a reference either.~ Turns out we can return a reference.

program/src/processor/mod.rs
```diff
@@ -144,3 +162,75 @@ macro_rules! require {
...
```
joncinque: My other PR didn't change this to ref, can this be
```suggestion
pub fn unpack_delegation_mut_uncheked(stake_data: &mut [u8]) -> Result<&mut Delegation, ProgramError> {
```

joncinque (commented):
Sorry one last bit, then this can go in

program/src/processor/mod.rs
febo: I removed some of the lifetimes that clippy said they could be inferred.

joncinque (approved):

PR #42, opened 2024-08-01
Processor: `InitializeSolStakerStake`

febo:
This PR adds the implementation of the new `InitializeSolStakerStake` instruction.


program/src/error.rs
```diff
@@ -88,6 +88,10 @@ pub enum StakeError {
     /// 19 - Invalid slash amount
     #[error("Invalid slash amount")]
     InvalidSlashAmount,
+
+    /// 20 - Undelegated stake state account
+    #[error("Undelegated stake state account")]
+    UndelegatedStakeStateAccount,
```
joncinque: nit: how about calling it a "SOL stake account"?
```suggestion
    /// 20 - Undelegated SOL stake account
    #[error("Undelegated SOL stake account")]
    UndelegatedSolStakeAccount,
```

program/src/instruction.rs
```diff
@@ -421,6 +421,49 @@ pub enum StakeInstruction {
...
+    #[account(
+        2,
+        writable,
+        name = "validator_stake",
+        desc = "Validator stake account (pda of `['stake::state::validator_stake', validator, config]`)"
+    )]
+    #[account(
+        3,
+        name = "stake_state",
+        desc = "SOL stake state account"
```
joncinque: nit: call it a SOL stake?
```suggestion
        name = "sol_stake",
        desc = "SOL stake account"
```

scripts/generate-clients.mjs
```diff
@@ -50,6 +50,13 @@ kinobi.update(
         "spl-token-2022"
       ),
     },
+    {
+      account: "stakeHistory",
+      defaultValue: k.publicKeyValueNode(
+        "SysvarStakeHistory1111111111111111111111111",
+        "stake-history"
+      ),
+    },
```
joncinque: nit: Is there a naming convention in place for sysvars? For example, if it's the clock, do you put it as `"clock"` or `"sysvarClock"`? If it's the former, then this is perfect, but if not, we should keep it consistent with other naming conventions for sysvars

program/src/processor/initialize_sol_staker_stake.rs
```diff
@@ -0,0 +1,187 @@
...
+/// There can be only one SOL staker stake account per stake state and config account, since
+/// the stake state is part of the SOL staker stake account seeds.
+///
+/// NOTE: Anybody can create the stake account for a SOL staker. For new
+/// accounts, the authority is initialized to the stake state account's withdrawer.
+///
+/// 0. `[]` Stake config account
+/// 1. `[w]` SOL staker stake account
+///     * PDA seeds: ['stake::state::sol_staker_stake', stake state, config]
+/// 2. `[]` SOL stake state account
```
joncinque: nit: this is missing the validator stake
```suggestion
/// 2. `[w]` Validator stake account
/// 3. `[]` SOL stake state account
```

program/src/processor/initialize_sol_staker_stake.rs
```diff
@@ -0,0 +1,187 @@
...
+        "config"
+    );
+
+    // stake state (validated by the SOL Stake View program)
+    // - must have a delegation
+
+    GetStakeActivatingAndDeactivatingCpiBuilder::new(ctx.accounts.sol_stake_view_program)
+        .stake(ctx.accounts.stake_state)
+        .stake_history(ctx.accounts.stake_history)
+        .invoke()?;
```
joncinque: Does this validate somewhere that the `sol_stake_view_program` is correct? That program is a huge attack target, so we should definitely be sure that we're calling the correct one!

clients/rust/tests/initialize_sol_staker_stake.rs
```diff
@@ -0,0 +1,450 @@
...
+        .stake_state(stake_state)
+        .sol_stake_view_program(paladin_sol_stake_view_program_client::ID)
+        .instruction();
+
+    let tx = Transaction::new_signed_with_payer(
+        &[initialize_ix],
+        Some(&context.payer.pubkey()),
+        &[&context.payer],
+        context.last_blockhash,
+    );
```
joncinque: Be careful, this can be flaky, since you're doing the same exact transaction twice, with the same signature. It's better to get a new blockhash before doing this.

clients/rust/tests/initialize_sol_staker_stake.rs
joncinque, line 1: Can you also add a test for a failure when providing the wrong sol stake view program id?

joncinque (commented):
Looks really good overall! Mostly small stuff

scripts/generate-clients.mjs
febo: I don't think we have a naming convention. Perhaps it would be a good idea to have the `sysvar_*` prefix to clearly indicate that it is a Sysvar account.

program/src/instruction.rs
febo: Should we also rename the field on the `SolStakerStake`? https://github.com/paladin-bladesmith/stake-program/blob/main/program/src/state/sol_staker_stake.rs#L40

joncinque: Ah yeah, please, maybe `sol_stake` or `sol_stake_address`? I'd lean towards the first one since we don't use `_address` anywhere else

program/src/processor/initialize_sol_staker_stake.rs
febo: Yes, it is indirectly validated since the CPI instruction is built with the address of the correct program. This is the error that you get when you pass the wrong program.
```bash
DEBUG Instruction: InitializeSolStakerStake
DEBUG Unknown program stkVUdWUiarMmkttUKMGLCLwHUkBqYfQ9vZcfG3T7LU
DEBUG Program PStake1111111111111111111111111111111111111 consumed 3037 of 399850 compute units
DEBUG Program PStake1111111111111111111111111111111111111 failed: An account required by the instruction is missing
```

But I will add a test so the error message is more clear.

clients/rust/tests/initialize_sol_staker_stake.rs
febo: Good point!

program/src/processor/initialize_sol_staker_stake.rs
febo: Done!
```
DEBUG Instruction: InitializeSolStakerStake
DEBUG Constraint failed: invalid sol stake view program
DEBUG Error: IncorrectProgramId
DEBUG Program PStake1111111111111111111111111111111111111 consumed 1778 of 399850 compute units
DEBUG Program PStake1111111111111111111111111111111111111 failed: incorrect program id for instruction
```

joncinque (approved):
Thanks for addressing all the comments!

PR #43, opened 2024-08-02
Refactor: `ValidatorStakeTokens`

febo:
This PR refactors the `StakeTokens` to `ValidatorStakeTokens` to reflect that it should only be used to stake tokens on `ValidatorStake` accounts.

It also introduces the limit on the total amount of tokens that can be staked in relation to the lamports (SOL) amount staked on the validator.


clients/rust/tests/validator_stake_tokens.rs
```diff
@@ -522,3 +531,199 @@ async fn fail_stake_tokens_with_uninitialized_stake_account() {
...
+
+    let mut stake_ix = ValidatorStakeTokensBuilder::new()
+        .config(config_manager.config)
+        .validator_stake(stake_manager.stake)
+        .source_token_account(rewards_manager.token_account)
+        .token_account_authority(rewards_manager.owner.pubkey())
+        .mint(config_manager.mint)
+        .vault(config_manager.vault)
+        .token_program(spl_token_2022::ID)
+        .amount(50_000_000_000) // <- stake 50 tokens (raw value with 9 digits precision)
```
joncinque: nit: it's typically more interesting to do the exact limits in tests, so in this case, we'd want to do `2_600_000_001` for it to just exactly fail

clients/rust/tests/validator_stake_tokens.rs
```diff
@@ -40,12 +41,20 @@ async fn stake_tokens() {
...
+    // Given a config account and a validator stake with 50 SOL staked.
 
     let config_manager = ConfigManager::new(&mut context).await;
     let stake_manager = ValidatorStakeManager::new(&mut context, &config_manager.config).await;
 
-    // And we initialize the holder rewards accounts and mint 100 tokens.
+    let mut account = get_account!(context, stake_manager.stake);
+    let mut stake_account = ValidatorStake::from_bytes(account.data.as_ref()).unwrap();
+    // "manually" set the amount to 50 SOL
+    stake_account.total_staked_lamports_amount = 50 * 1_000_000_000;
```
joncinque: nit: How about setting this or the staked PAL amount to exactly the limit allowable for the validator? It should be 77 lamports if I got the rounding correct?

program/src/error.rs
```diff
@@ -88,6 +88,10 @@ pub enum StakeError {
     /// 19 - Invalid slash amount
     #[error("Invalid slash amount")]
     InvalidSlashAmount,
+
+    /// 20 - Total stake amount exceeds limit
+    #[error("Total stake amount exceeds limit")]
+    TotalStakeAmountExceedsLimit,
```
joncinque: nit: Maybe this can be clearer, like "Total PAL stake amount exceeds 1.3 * SOL stake amount", with name `TotalPalStakeAmountExceedsLimit`. What do you think?

program/src/processor/validator_stake_tokens.rs
```diff
@@ -114,25 +110,32 @@ pub fn process_stake_tokens<'a>(
     let decimals = mint.base.decimals;
 
     // Update the config and stake account data.
-    //
-    // Note: validate the amount against the total SOL staked on the validator.
 
     require!(amount > 0, StakeError::InvalidAmount);
 
-    require!(amount > 0, StakeError::InvalidAmount);
```
joncinque, line 122: What happened here in the old code? :sweat_smile:

joncinque (commented):
Just some nits, looks great!

program/src/error.rs
febo: It definitely looks better, but I wonder:
1. We haven't made direct references to PAL token in the program so far and technically you could use it with other tokens.
2. the `1.3` multiplier might end up not being fixed, so I wonder if it is a good idea to have it on an error message.

A bit related to this, I was going to ask whether we should consider renaming the `total_staked_pal_amount` field to something more generic or not – like `total_staked_token_amount` – similar to how the name is on the `Config`.

https://github.com/paladin-bladesmith/stake-program/blob/main/program/src/state/validator_stake.rs#L36

program/src/processor/validator_stake_tokens.rs
febo: I was equally surprised. 😅 I think this is a result of having the same commit in separate PRs due to "squash and merge" that we discussed the other day.

clients/rust/tests/validator_stake_tokens.rs
febo: Nice, that is a really good point, thanks!

clients/rust/tests/validator_stake_tokens.rs
febo: Good point!

program/src/error.rs
joncinque: Ah gotcha, that's a good point. Feel free to rename if you think it's clearer. In that case though, maybe we can call the error `TotalStakeAmountExceedsSolLimit` :smile:

joncinque (approved):
Looks great!

PR #46, opened 2024-08-06
Processor: `SyncSolStake`

febo:
This PR adds the implementation for the `SyncSolStake` instruction.


program/src/instruction.rs
```diff
@@ -683,6 +721,8 @@ impl StakeInstruction {
 
                 Ok(StakeInstruction::SolStakerStakeTokens(amount))
             }
+            // 13 - InitializeSolStakerStake
```
joncinque: micro-nit
```suggestion
            // 14 - SyncSolStake
```

program/src/processor/sync_sol_stake.rs
```diff
@@ -0,0 +1,162 @@
...
+    state::{
+        find_sol_staker_stake_pda, find_validator_stake_pda, Config, SolStakerStake, ValidatorStake,
+    },
+};
+
+/// Sync the SOL stake balance with a validator and SOL staker stake accounts.
+///
+/// NOTE: Anybody can sync the balance of a SOL stake account.
+///
+/// 0. `[w]` SOL staker stake account
```
joncinque: nit: this list is missing the stake config account

program/src/processor/sync_sol_stake.rs
```diff
@@ -0,0 +1,162 @@
...
+    //   3. Add the new stake amount to the validator stake account.
+
+    validator_stake.total_staked_lamports_amount = validator_stake
+        .total_staked_lamports_amount
+        .checked_sub(sol_staker_stake.lamports_amount)
+        .ok_or(ProgramError::ArithmeticOverflow)?;
+
+    sol_staker_stake.lamports_amount = u64::from(stake_state_data.activating)
+        .checked_add(stake_state_data.effective.into())
+        .ok_or(ProgramError::ArithmeticOverflow)?;
```
joncinque: If we're including `activating` in the `lamports_amount`, then we should subtract `deactivating`. I think we'll need to do this, because if we don't sync when it's deactivating, we might miss it totally if the user immediately activates onto another validator after the deactivation is complete.

clients/rust/tests/sync_sol_stake.rs
```diff
@@ -0,0 +1,637 @@
...
+    );
+
+    // And we deactivate the stake.
+
+    deactivate_stake_account(
+        &mut context,
+        &stake_account.sol_stake,
+        &sol_staker_staker_manager.authority,
+    )
+    .await;
```
joncinque, line 80: This test hasn't warped to stake activation time, so the deactivation is instantaneous, and it didn't catch the `deactivating` issue brought up in the processor. Can you add a warp to make sure that the stake is active before deactivating it?

joncinque (commented):
Looks really great overall! Just the one bit about `deactivating`

program/src/processor/sync_sol_stake.rs
febo: Cool, I will update to subtract `deactivating`.

febo: I wonder if we need to improve the logic further. Is there a possibility that we miss to sync while the stake is deactivating (e.g., the off-chain component does not run) and the stake gets activated with another validator? The "pointer" to the sol stake would still be valid, since it would be equal to the value stored on the account, so this will pass:
```rust
require!(
    ctx.accounts.sol_stake.key == &sol_staker_stake.sol_stake,
    StakeError::IncorrectSolStakeAccount,
    "sol stake"
);
```
But if we add a test for the `delegated_vote` it will missmatch since the account is delegated to a different validator.

So perhaps if we encounter this scenario we need to update the delegation to `0`.

febo: Improved the sync logic + added a test to simulate this scenario:
* [Improve sync logic](https://github.com/paladin-bladesmith/stake-program/pull/46/commits/b6f917d2251d192e252f9a09c872d11226c8305f)

joncinque (approved):
Looks great, and nice find on the extra edge case!

PR #47, opened 2024-08-07
Refactor: `HarvestValidatorRewards`

febo:
This PR refactors the `HarvestStakeRewards` instruction into `HarvestValidatorRewards`. This instruction is used to harvest the rewards of `ValidatorStake` stakers.


clients/rust/tests/harvest_validator_rewards.rs
```diff
@@ -203,85 +203,54 @@ async fn harvest_stake_rewards_after_harvesting() {
...
 
-    let mut account = get_account!(context, stake_pda);
+    let mut account = get_account!(context, validator_stake_manager.stake);
     let mut stake_account = ValidatorStake::from_bytes(account.data.as_ref()).unwrap();
     // "manually" set the amount to 50
-    stake_account.amount = 50;
+    stake_account.delegation.amount = 50;
+    stake_account.total_staked_lamports_amount = 50;
+    // 0.04 * 1e9
+    stake_account.delegation.last_seen_stake_rewards_per_token = 40_000_000 * 1_000_000_000;
```
joncinque: This is a bit confusing -- unless I'm misunderstanding the concept of the test, how about just setting it to `config_account.accumulated_stake_rewards_per_token` since the idea is that everything has been harvested already?

clients/rust/tests/harvest_validator_rewards.rs
```diff
@@ -45,81 +43,86 @@ async fn harvest_stake_rewards() {
...
     // Calculation:
-    //   - total staked: 100
-    //   - stake rewards: 4 SOL
-    //   - rewards per token: 4_000_000_000 / 100 = 40_000_000 (0.04 SOL)
-    //   - rewards for 50 staked: 40_000_000 * 50 = 2_000_000_000 (2 SOL)
+    //   - total staked: 130
+    //   - stake rewards: 26 lamports
+    //   - rewards per token: 26 / 130 = 0.2
+    //   - validator stake amount: 65 (limit 1.3 * 50 = 65)
+    //   - rewards for 65 staked: 0.2 * 65 = 13 lamports
```
joncinque, line 83: This is great!

clients/rust/tests/harvest_validator_rewards.rs
```diff
@@ -579,3 +510,119 @@ async fn fail_harvest_stake_rewards_with_wrong_config_account() {
...
+    //   - the excess rewards are 3_250_000_000 (3.25 SOL), which goes "back" to
+    //     the config account: 6.5 + 3.25 =  9_750_000_000 (9.75 SOL)
+    //
+    //   - and the excess is shared only with the remaining staked amount:
+    //     26_000_000_000 - 13_000_000_000 = 13_000_000_000
+    //
+    //   - the rewards per token for the remaining staked amount:
+    //     3_250_000_000 / 13_000_000_000 = 0.25 SOL
+    //
+    //   - the final accumulated stake rewards per token: 750_000_000 (0.5 + 0.25 SOL)
```
joncinque, line 585: Thanks for laying this all out, makes sense to me!

joncinque (approved):
Looks really good! Just one little bit on the test, but feel free to merge if you want

program/src/processor/harvest_validator_rewards.rs
```diff
@@ -0,0 +1,205 @@
...
+    // The rewards are caped using the minimum of the total amount staked and the
+    // limit calculated from the total SOL amount staked. This is to prevent getting
+    // rewards from exceeding token amount based on the total SOL amount staked.
+    //
+    // Any excess rewards are distributes back to stakers and the validator forfeits
+    // its share of them.
+
+    let stake_limit =
+        calculate_maximum_stake_for_lamports_amount(validator_stake.total_staked_lamports_amount)?;
+    let validated_amount = min(validator_stake.delegation.amount, stake_limit as u64);
```
buffalojoec, line 108: This should be okay, but it seems to me that it would be safer to cast the stake limit to `u64` within the helper function as fallible (ie. `try_into()`).

In a very extreme example, a stake amount of `(u64::MAX / 1.3) + 1` would break this cast.

It's typically best to avoid blindly casting a retrieved `u128` to `u64`, but since we're extremely unlikely to have this amount of stake it's really just a semantics recommendation.

program/src/processor/harvest_validator_rewards.rs
```diff
@@ -0,0 +1,205 @@
...
+                stake_limit
+            );
+
+            let excess = (validator_stake.delegation.amount as u128)
+                .checked_sub(stake_limit)
+                .ok_or(ProgramError::ArithmeticOverflow)?;
+
+            let excess_rewards = calculate_eligible_rewards(
+                accumulated_rewards_per_token,
+                u128::from(validator_stake.delegation.last_seen_stake_rewards_per_token),
```
buffalojoec: micro-nit: Looks like you do this conversion twice. Could maybe use a variable like you do with `accumulated_rewards_per_token`?

program/src/processor/harvest_validator_rewards.rs
```diff
@@ -0,0 +1,205 @@
...
+                config
+                    .token_amount_delegated
+                    .checked_sub(validator_stake.delegation.amount)
+                    .ok_or(ProgramError::ArithmeticOverflow)?,
+            )?;
+
+            if rewards_per_token != 0 {
+                // Updates the accumulated stake rewards per token on the config to
+                // reflect the addition of the rewards for the exceeding amount.
+                let accumulated = u128::from(config.accumulated_stake_rewards_per_token)
```
buffalojoec: nit: Same here, you've already got this as a variable `accumulated_rewards_per_token`.

program/src/processor/harvest_validator_rewards.rs
```diff
@@ -0,0 +1,205 @@
...
+            );
+
+            let excess = (validator_stake.delegation.amount as u128)
+                .checked_sub(stake_limit)
+                .ok_or(ProgramError::ArithmeticOverflow)?;
+
+            let excess_rewards = calculate_eligible_rewards(
+                accumulated_rewards_per_token,
+                u128::from(validator_stake.delegation.last_seen_stake_rewards_per_token),
+                excess as u64,
```
buffalojoec, line 178: Same thing here with the `u64` cast comment. You wouldn't have to repeat this if the helper just returned it. Looks like everywhere you use this helper you either cast to `u64` or cast a `u64` up to `u128` to compare.

clients/rust/tests/harvest_validator_rewards.rs
```diff
@@ -203,85 +203,54 @@ async fn harvest_stake_rewards_after_harvesting() {
...
     context.banks_client.process_transaction(tx).await.unwrap();
 
     // Then the destination account has no rewards.
 
-    let account = context
-        .banks_client
-        .get_account(second_destination)
-        .await
-        .unwrap();
+    let account = context.banks_client.get_account(destination).await.unwrap();
```
buffalojoec, line 254: You might want to fund the destination as rent-exempt here just to be sure, since the reward amount could be below rent-exemption and the account could be garbage collected before you run this check. Maybe instead fund it and ensure the balance didn't change?

buffalojoec (commented):
This looks really good, nice work! I left a couple of micro-nits, but I've also left a suggestion about integer casting, if you were interested in refactoring those helpers.

clients/rust/tests/harvest_validator_rewards.rs
febo: Yep, you are right.

program/src/processor/harvest_validator_rewards.rs
febo: Nice, I will add the `try_into()`.

clients/rust/tests/harvest_validator_rewards.rs
febo: Here the default behaviour works. When you add lamports to an account that do not cover the rent-exemption, you get a `TransactionError::InsufficientFundsForRent`.

buffalojoec: Ahh, cool, perfect!

buffalojoec (approved):

PR #49, opened 2024-08-09
Processor: `HarvestSyncRewards`

febo:
This PR adds the processor for the `HarvestSyncRewards` instructions. As part of it, the `Config` has a new field to set the `sync_rewards_lamports`, which is the rewards given for syncing the SOL stake amount.


program/src/processor/harvest_sync_rewards.rs
```diff
@@ -0,0 +1,232 @@
...
+        || delegated_vote != Some(sol_staker_stake.delegation.validator_vote)
+    {
+        stake_state_data = bytemuck::from_bytes(&EMPTY_RETURN_DATA);
+    }
+
+    // Check whether the SOL stake is ouf-of-sync.
+
+    let stake_amount = u64::from(stake_state_data.effective)
+        .checked_add(u64::from(stake_state_data.activating))
+        .ok_or(ProgramError::ArithmeticOverflow)?;
```
joncinque: I believe this should remove the deactivating amount too

program/src/processor/harvest_sync_rewards.rs
```diff
@@ -0,0 +1,232 @@
...
+
+        // searcher rewards
+        let rewards = calculate_eligible_rewards(
+            u128::from(config.accumulated_stake_rewards_per_token),
+            u128::from(
+                sol_staker_stake
+                    .delegation
+                    .last_seen_stake_rewards_per_token,
+            ),
+            sol_staker_stake.delegation.amount,
```
joncinque, line 182: Shouldn't this use `min(sol_staker_stake.delegation.amount, stake_amount)`?

program/src/processor/harvest_sync_rewards.rs
```diff
@@ -0,0 +1,232 @@
...
+
+        if searcher_rewards > 0 {
+            // update last seen stake rewards per token
+            sol_staker_stake
+                .delegation
+                .last_seen_stake_rewards_per_token = calculate_stake_rewards_per_token(
+                searcher_rewards,
+                sol_staker_stake.delegation.amount,
+            )?
+            .into();
```
joncinque: Shouldn't this *add* `calculate_stake_rewards_per_token` to `last_seen_stake_rewards_per_token`?

program/src/processor/harvest_sync_rewards.rs
```diff
@@ -0,0 +1,232 @@
...
+        )?;
+        let searcher_rewards = min(config.sync_rewards_lamports, rewards);
+
+        if searcher_rewards > 0 {
+            // update last seen stake rewards per token
+            sol_staker_stake
+                .delegation
+                .last_seen_stake_rewards_per_token = calculate_stake_rewards_per_token(
+                searcher_rewards,
+                sol_staker_stake.delegation.amount,
```
joncinque: Same here, I think this should be `min(sol_staker_stake.delegation.amount, stake_amount)`

clients/rust/tests/harvest_sync_rewards.rs
```diff
@@ -0,0 +1,905 @@
...
+    //   right: 999230769
+    /*
+    assert_eq!(
+        config_account
+            .accumulated_stake_rewards_per_token
+            .checked_sub(last_seen_stake_rewards_per_token)
+            .unwrap(),
+        calculate_stake_rewards_per_token(1_299_000_000, 1_300_000_000)
+    );
+    */
```
joncinque: This definitely is important -- at the very least, things need to be consistent, but I hope that this will be resolved with the other parts that I brought up

clients/rust/tests/harvest_sync_rewards.rs
```diff
@@ -0,0 +1,905 @@
...
+    );
+
+    // And we close the SOL stake account.
+
+    context.set_account(
+        &stake_account.sol_stake,
+        &AccountSharedData::from(Account {
+            data: vec![],
+            lamports: 0,
+            owner: system_program::ID,
```
joncinque: You might want to check this once locally, but it's possible withdraw everything from a stake account and then use it again within the same transaction because the stake program doesn't reallocate to 0 / set the owner to the system program.

So what happens if the owner is the stake program and there's no data?

joncinque (commented):
A few comments, but the guts look good

program/src/processor/harvest_sync_rewards.rs
febo: Yep, you are right.

program/src/processor/harvest_sync_rewards.rs
febo: Absolutely, nice catch! I will add a test for this.

program/src/processor/harvest_sync_rewards.rs
febo: We are not capping the rewards here, so we have funds to pay the sync rewards. In a scenario where `stake_amount` is `0` but there are rewards available, the sync rewards are paid from the excess.

program/src/processor/harvest_sync_rewards.rs
febo: Same as the previous comment, sync rewards are paid from the excess.

program/src/processor/harvest_sync_rewards.rs
febo: Added the following to the `harvest_sync_rewards_with_capped_sync_rewards` test:
```rust
// add 200_000_000 (0.2) to the accumulated stake rewards per token to simulate
// a previous reward
config_account.accumulated_stake_rewards_per_token =
        calculate_stake_rewards_per_token(1_300_000_000, 1_300_000_000) + 200_000_000;
```
So the following assert tests that the rewards  per token have been accumulated on the stake account:
```rust
assert_eq!(
    config_account.accumulated_stake_rewards_per_token,
    stake_account.delegation.last_seen_stake_rewards_per_token
);
```

clients/rust/tests/harvest_sync_rewards.rs
febo: Works the same way. Should I keep `system_program::ID` or the stake program on the test, or create a new test with the stake program?

joncinque: Ok great! No worries, feel free to keep this

program/src/processor/harvest_sync_rewards.rs
joncinque: There's some weird game theory here then. You can keep your PAL account over-staked (PAL amount much higher than SOL) and use this to get extra rewards that you otherwise wouldn't get if you did normal syncing.

joncinque: But I think you're right, this probably makes the most sense -- oh well!

joncinque (approved):
Looks great!

PR #51, opened 2024-08-09
Refactor: `UpdateConfig`

febo:
### Problem

The `sync_rewards_lamports` field was added to `Config` as part of the `HarvestSyncRewards` instruction, but it is not currently possible to update its value after the creation of the `Config`.

### Solution

This PR refactors the `UpdateConfig` processor to support updating the value of `sync_rewards_lamports`. It also adds a test for it.

### DIFF
--- a/clients/rust/tests/update_config.rs
+++ b/clients/rust/tests/update_config.rs
@@ -220,6 +220,107 @@ async fn update_max_deactivation_basis_points_config() {
 }
 
 #[tokio::test]
+async fn update_sync_rewards_lamports() {
+    let mut context = ProgramTest::new(
+        "paladin_stake_program",
+        paladin_stake_program_client::ID,
+        None,
+    )
+    .start_with_context()
+    .await;
+
+    // Given an empty config account and a mint.
+
+    let config = Keypair::new();
+    let authority = Keypair::new();
+
+    let mint = Keypair::new();
+    create_mint(
+        &mut context,
+        &mint,
+        &authority.pubkey(),
+        Some(&authority.pubkey()),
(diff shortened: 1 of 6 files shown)


clients/rust/tests/update_config.rs
```diff
@@ -219,6 +219,107 @@ async fn update_max_deactivation_basis_points_config() {
...
+        &[&context.payer, &config],
+        context.last_blockhash,
+    );
+    context.banks_client.process_transaction(tx).await.unwrap();
+
+    let account = get_account!(context, config.pubkey());
+    let config_account = Config::from_bytes(account.data.as_ref()).unwrap();
+    assert_eq!(config_account.sync_rewards_lamports, 1_000_000);
+
+    // When we update the cooldown time config.
```
joncinque:
```suggestion
    // When we update the sync rewards lamports
```

clients/rust/tests/update_config.rs
```diff
@@ -219,6 +219,107 @@ async fn update_max_deactivation_basis_points_config() {
...
+
+    let tx = Transaction::new_signed_with_payer(
+        &[ix],
+        Some(&context.payer.pubkey()),
+        &[&context.payer, &authority],
+        context.last_blockhash,
+    );
+    context.banks_client.process_transaction(tx).await.unwrap();
+
+    // Then the cooldown time was updated.
```
joncinque:
```suggestion
    // Then the sync rewards lamports field was updated.
```

joncinque (approved):
Just some comments that used another field name, looks good otherwise!

PR #56, opened 2024-08-12
Refactor: `InactivateValidatorStake`

febo:
### Problem

Currently there is a single `InactivateStake` instruction. Given that now there are two different types of stake accounts and the logic differs, the processor implementation needs to be refactored.

### Solution

Repurpose the current instruction for `InactivateValidatorStake`, which only inactivate stake from `ValidatorStake` accounts. A new instruction specific for `SolStakerStake` accounts will be added in a separate PR.


joncinque (approved):
It's nice when they're simple :-D

PR #57, opened 2024-08-12
Processor: `InactivateSolStakerStake`

febo:
### Problem

As part of the refactoring of `InactivateValidatorStake`, there is no way to inactivate stake held in a `SolStakerStake` account.

### Solution

This PR adds the processor implementation for a new `InactivateSolStakerStake` instruction.


joncinque (approved):
Looks great!

PR #65, opened 2024-08-22
Processor: Increase rewards per token precision

buffalojoec:
#### Problem
With the program's current scaling factor for "rewards per token" of `1e9`, the minimum reward that will not cause `calculate_rewards_per_token` to return zero is 1 SOL. This is much too high, since small rewards could be paid into the system periodically.

#### Summary of Changes
Increase the scaling factor to `1e18`. This might be considered overkill, but our system can support it, and it also allows the program to capture rewards per token calculations for a minimum reward of 1 lamport.

For more information, see https://github.com/paladin-bladesmith/rewards-program/pull/29.

---

Increasing the scaling to `1e18` does reduce the upper bound of the maximum marginal rate, since within `calculate_eligible_rewards`, we are multiplying the marginal rate by the token account balance. This will be handled in a follow-up PR, where we'll change the checked math to wrapped math.

### DIFF
--- a/clients/rust/tests/setup/mod.rs
+++ b/clients/rust/tests/setup/mod.rs
@@ -15,8 +15,8 @@ pub mod token;
 pub mod validator_stake;
 pub mod vote;
 
-/// Scaling factor for rewards per token (1e9).
-pub const REWARDS_PER_TOKEN_SCALING_FACTOR: u128 = 1_000_000_000;
+/// Scaling factor for rewards per token (1e18).
+pub const REWARDS_PER_TOKEN_SCALING_FACTOR: u128 = 1_000_000_000_000_000_000;
 
 pub fn new_program_test() -> ProgramTest {
     let mut program_test = ProgramTest::new(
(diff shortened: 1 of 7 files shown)


clients/rust/tests/setup/mod.rs
```diff
@@ -141,9 +141,9 @@ pub fn calculate_stake_rewards_per_token(rewards: u64, stake_amount: u64) -> u12
     } else {
         // Calculation: rewards / stake_amount
         //
-        // Scaled by 1e9 to store 9 decimal places of precision.
+        // Scaled by 1e18 to store 18 decimal places of precision.
         (rewards as u128)
-            .checked_mul(1_000_000_000)
+            .checked_mul(1_000_000_000_000_000_000)
```
joncinque: nit: should we use the constant here?
```suggestion
            .checked_mul(REWARDS_PER_TOKEN_SCALING_FACTOR)
```

joncinque (approved):
Looks great!

clients/rust/tests/setup/mod.rs
buffalojoec: Yeah sure!

joncinque (approved):

PR #66, opened 2024-08-23
Processor: Use wrapped math

buffalojoec:
#### Problem
Whenever any reward system's "accumulated rewards per token" reaches a value of u128::MAX, that system would break and no longer accept any additional rewards paid in. Since we've increased the scaling factor to 1e18, the total system-level rewards paid in over time that would break that system is u128::MAX.

Although this is a pretty large upper bound, we can further increase each system's robustness by using wrapped math for incrementing "accumulated rewards per token" as well as calculating eligible rewards.

#### Summary of Changes
Use wrapped math instead of checked math for incrementing "accumulated rewards per token" in both `DistributeRewards` and `HarvestSyncRewards`, as well as calculating eligible rewards.

### DIFF
--- a/clients/rust/tests/distribute_rewards.rs
+++ b/clients/rust/tests/distribute_rewards.rs
@@ -73,6 +73,67 @@ async fn distribute_rewards() {
 }
 
 #[tokio::test]
+async fn distribute_rewards_wrapped() {
+    let mut context = ProgramTest::new(
+        "paladin_stake_program",
+        paladin_stake_program_client::ID,
+        None,
+    )
+    .start_with_context()
+    .await;
+
+    // Given a config account.
+
+    let config = create_config(&mut context).await;
+    // "manually" set the total amount delegated and max reward.
+    let account = get_account!(context, config);
+    let mut config_account = Config::from_bytes(account.data.as_ref()).unwrap();
+    config_account.token_amount_delegated = 5; // <- 5 tokens
+    config_account.accumulated_stake_rewards_per_token = u128::MAX; // Maximum rate.
+
+    let updated_config = Account {
+        lamports: account.lamports,
+        data: config_account.try_to_vec().unwrap(),
+        owner: account.owner,
+        executable: account.executable,
+        rent_epoch: account.rent_epoch,
+    };
+    context.set_account(&config, &updated_config.into());
+
+    // When we distribute rewards.
+
+    let distribute_rewards_ix = DistributeRewardsBuilder::new()
+        .config(config)
+        .payer(context.payer.pubkey())
+        .amount(5_000_000_000)
+        .instruction();
+
+    let tx = Transaction::new_signed_with_payer(
+        &[distribute_rewards_ix],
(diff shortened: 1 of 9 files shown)


clients/rust/tests/distribute_rewards.rs
```diff
@@ -72,6 +72,67 @@ async fn distribute_rewards() {
     );
 }
 
+#[tokio::test]
```
buffalojoec, line 75: Note: All of these "wrapped" tests were created by simply copy-pasting the success case and then configuring a wrapping reward rate on the system.

program/src/state/mod.rs
```diff
@@ -309,4 +308,30 @@ mod tests {
...
+        // current reward has wrapped around `u128::MAX`.
+        let current_accumulated_rewards_per_token = 1_000_000_000_000_000_000;
+        let last_accumulated_rewards_per_token = u128::MAX - 1_000_000_000_000_000_000;
+        let result = calculate_eligible_rewards(
+            current_accumulated_rewards_per_token,
+            last_accumulated_rewards_per_token,
+            BENCH_TOKEN_SUPPLY,
+        )
+        .unwrap();
+        assert_ne!(result, 1_000_000_000_000_000_000);
```
joncinque: nit: I believe this was lifted from the other test, but can it test for an exact amount?

clients/rust/tests/harvest_sync_rewards.rs
```diff
@@ -197,6 +197,176 @@ async fn harvest_sync_rewards() {
...
+        .accumulated_stake_rewards_per_token
+        .checked_sub(last_seen_stake_rewards_per_token)
+        .and_then(|rewards| {
+            rewards.checked_sub(calculate_stake_rewards_per_token(
+                1_299_000_000,
+                1_300_000_000,
+            ))
+        })
+        .unwrap()
+        - 1;
```
joncinque: micro-nit: the `-1` was a bit confusing, maybe it's clearer as 
```suggestion
    let difference = config_account
        .accumulated_stake_rewards_per_token
        .checked_sub(last_seen_stake_rewards_per_token)
        .and_then(|rewards| {
            rewards.checked_sub(calculate_stake_rewards_per_token(
                1_299_000_000,
                1_300_000_000,
            ) - 1)
        })
        .unwrap();
```

What do you think?

joncinque (commented):
Looks good! Just a few little things

We also need to update the harvesting logic at https://github.com/paladin-bladesmith/stake-program/blob/4eeb7495b104be484242108ff932bbf8b3ceca69/program/src/processor/mod.rs#L409-L414

clients/rust/tests/harvest_sync_rewards.rs
buffalojoec: Sure, if you think that's more clear I'll add the change, but that would become `+ 1` instead.

joncinque: that works, thanks!

joncinque (approved):

PR #90, opened 2025-07-28
Token 2022 removal

Art3miX:
There are 2 big changes that required changes for almost all tests.

1. Create `HolderRewards` account from within the initializeConfig instruction, this is because the rewards program was changes to require the owner of the token account to sign the transaction, and the owner of the stake vault is a PDA.
2. Remove token2022 and transfer hooks.


program/src/processor/mod.rs
```diff
@@ -516,6 +517,33 @@ fn process_slash_for_delegation(args: SlashArgs) -> ProgramResult {
...
+    let source_data_len = source_account_info.data_len();
+    let min_rent = rent.minimum_balance(source_data_len);
+
+    // Ensure source will still be rent-exempt after transfer
+    if source_lamports < min_rent {
+        return Err(ProgramError::InsufficientFunds);
+    }
+
+    // Excess is what we currently have minus rent
+    let excess_lamports = source_lamports - min_rent;
```
OliverNChalk:
```suggestion
    let excess_lamports = source_lamports.checked_sub(min_rent).ok_or(ProgramError::InsufficientFunds);
```

program/src/processor/sol_staker_stake_tokens.rs
```diff
@@ -120,17 +125,25 @@ pub fn process_sol_staker_stake_tokens<'a>(
...
+        &[
+            ctx.accounts.source_token_account.clone(),
+            ctx.accounts.mint.clone(),
+            ctx.accounts.vault.clone(),
+            ctx.accounts.source_token_account_authority.clone(),
+        ],
     )?;
 
+    // TODO: Deposit tokens into holder rewards
     Ok(())
```
OliverNChalk, line 148: This will be in a subsequent PR?

Art3miX: Correct, this PR is already huge, I will work on the deposit, withdraw and harvest holder rewards next, which is the only missing piece in the stake program.

